PR #2, opened 2021-07-30
Plio usecase

sudheshna-donthineni:
* Please ignore the `StudentSelect.vue` file. I did try to not include it in the commit but it came up anyway. Changes to the code and renaming variables brought about a shift in behaviour in the dropdown. I have to anyway nix the dummy file and load real data so I tried to not put effort into that right now as the plio use case is time-sensitive. 

- Sometimes I added multi-line comments to communicate the logic. Once we have a documentation process ready, I think we can transfer these comments and reduce them to single-line.

- Reiterating, for the plio use case, multiple SRN entries are disabled. Because plio has a singular unique ID param. This could be temporary once we figure out the plio backend of accepting multiple ID's but producing a single link. 

- The way I have tested it is by manually adding the params to the url.


public/index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <link rel="icon" href="<%= BASE_URL %>favicon.ico">
```
deepansh96: change the favicon to AF's logo

src/assets/list.js
```diff
@@ -0,0 +1,67 @@
+const listOfSchools = [
```
deepansh96: This file is not needed for this PR right?
You can keep it in your local but I would suggest not merging it into master

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,106 @@
...
+
+<script>
+export default {
+  name: "SRNEntry",
+  props: {
+    redirectTo: String,
+    redirectID: String,
+    purpose: String,
+    subPurpose: String,
+    plioUse: Boolean,
```
pritamps: What is `plioUse` for?

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,106 @@
...
+  },
+  data() {
+    return {
+      userIDList: [{ userID: "" }],
+      singleUserID: "", //is a temp hack
+    };
+  },
+  computed: {
+    buttonStyleClass() {
+      return "block bg-primary hover:bg-primary-hover text-white uppercase text-lg mx-auto p-4 mt-4 rounded";
```
pritamps: I still don't get why this is better than using `@apply` in the styles :D @deepansh96 thoughts?

src/components/StudentSelect.vue
```diff
@@ -0,0 +1,70 @@
+<template>
```
deepansh96: Again, this file is not needed for this PR (as you said to ignore it).
Best to keep it in your local but not commit it.

Otherwise after the review of this PR, this file might get merged into the main branch without being reviewed or tested

Your other PR, `firstauth-layer` is the place where this file should live and after review/testing should get merged to the main branch

If you need it as a placeholder, you can just commit an empty file with this name without any code

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,106 @@
...
+    },
+    labelStyleClass() {
+      return "mb-2 mx-auto uppercase font-bold text-lg text-grey-darkest";
+    },
+    inputStyleClass() {
+      return "border py-2 mx-auto px-3 text-grey-darkest";
+    },
+    isAnyUserIDPresent() {
+      return this.userIDList.length > 1;
```
pritamps: Especially because now styles will be mixed with other computed properties

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,106 @@
...
+    buttonStyleClass() {
+      return "block bg-primary hover:bg-primary-hover text-white uppercase text-lg mx-auto p-4 mt-4 rounded";
+    },
+    labelStyleClass() {
+      return "mb-2 mx-auto uppercase font-bold text-lg text-grey-darkest";
+    },
+    inputStyleClass() {
+      return "border py-2 mx-auto px-3 text-grey-darkest";
+    },
+    isAnyUserIDPresent() {
```
pritamps: Why not just check if the `userIdList` is empty?

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,106 @@
...
+      return this.userIDList.length > 1;
+    },
+  },
+  methods: {
+    //checking to see if each char typed by user is only a letter or number
+    isNumberOrLetter(e) {
```
pritamps: Rename this to `isValidSRN` because that is what we are really checking, no?

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+    <h3 class="mx-auto">OR</h3>
+    <StudentSelect />
+  </div>
+</template>
+
+<script>
+import StudentSelect from "@/components/StudentSelect.vue";
+import SRNEntry from "@/components/SRNEntry.vue";
+
+/* Have a variable plioUse that will help toggle between divs, one that will disable the entry of more
```
pritamps: Hmm... I don't see this design extending beyond this. Whether multiple SRNs are allowed or not doesn't necessarily depend on whether its `plioUse` or not, so the two checks should be separate.  Since this variable is *only* checking if multiple SRNs shoudl be used or not, might as well rename to `multipleSRNsAllowed"

The auth should get all these things as params.

A better design might be to have separate components for a single-select SRN view vs a multi-SRN view, both of which will use the same SRN input component.

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+export default {
+  name: "HomePage",
+  components: {
+    StudentSelect,
+    SRNEntry,
+  },
+
+  /*these are props expected by this app. 
+  RedirectTo tells what is the resource we are redirecting to. Eg. redirectTo=plio tells us that we are redirecting to a plio. 
+  The plio link is built later on after the student submits their SRN.
```
pritamps: No reference to plio link here in these docs. Here we don't know what the purpose is, yet

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+  RedirectID contains the ID of the resource. Eg. the plioID
+  Purpose is the general category of why the data is being captured. Eg: attendance
+  SubPurpose describes the subcategory of the purpose. Eg: plio -> means the attendance is for a plio link
```
pritamps: Here plio makes sense becuase it's used only as an example :)

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+    },
+    redirectID: {
+      default: "",
+      type: String,
+    },
+    purpose: {
+      default: "",
+      type: String,
+    },
+    subPurpose: {
```
pritamps: Maybe just rename this to `params`?

pritamps (changes requested):
Just some design comments. Will let @deepansh96 comment in more detail

src/pages/Homepage.vue
sudheshna-donthineni: I did initially do that; created two components: SingleSRNEntry, SRNEntry. But then I realized that a lot of code was repeating, the only thing that is different is, if it allows multiple entries, the var is of an object type, otherwise, it's a string. The plio code takes the `unique_id` as a string, so that was when I decided to just have one component and toggle between two variables.

sudheshna-donthineni: But yes, instead of hardcoding the `plioUse` var, I could check if the `Purpose` is `attendance` and the `subPurpose` is `plio`, then use one variable over the other.

src/pages/Homepage.vue
sudheshna-donthineni: I did not understand this, @pritamps :(

src/components/SRNEntry.vue
sudheshna-donthineni: I've changed it to `isValidSRNFormat`. I will be using a function called `isValidSRN` when the data is fetched and compared from the backend.

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <SRNEntry
+    :redirectTo="redirectTo"
+    :redirectID="redirectID"
+    :purpose="purpose"
+    :subPurpose="subPurpose"
+    :plioUse="plioUse"
+  />
+  <div v-if="!plioUse">
```
deepansh96:
```suggestion
  <div v-if="!isRedirectToPlio">
```

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+    SRNEntry,
+  },
+
+  /*these are props expected by this app. 
+  RedirectTo tells what is the resource we are redirecting to. Eg. redirectTo=plio tells us that we are redirecting to a plio. 
+  The plio link is built later on after the student submits their SRN.
+  RedirectID contains the ID of the resource. Eg. the plioID
+  Purpose is the general category of why the data is being captured. Eg: attendance
+  SubPurpose describes the subcategory of the purpose. Eg: plio -> means the attendance is for a plio link
+*/
```
deepansh96: The best way we've found for putting comments to explain props is to put it inside the prop definition itself.
See my comment below

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+      type: String,
+    },
+    purpose: {
+      default: "",
+      type: String,
+    },
+    subPurpose: {
+      default: "",
+      type: String,
+    },
```
deepansh96:
```suggestion
    redirectTo: {
      // resource to redirect to
      default: "",
      type: String,
    },
    redirectID: {
      // ID of the resource to redirect to
      default: "",
      type: String,
    },
    purpose: {
      // purpose for data capturing
      default: "",
      type: String,
    },
    subPurpose: {
      // sub cateogry of purpose for data capturing
      default: "",
      type: String,
    },
```
You can add examples here as well

src/pages/Homepage.vue
```diff
@@ -0,0 +1,64 @@
...
+  },
+  data() {
+    return {
+      plioUse: true,
```
deepansh96: Discussed above as well,  best to write the logic to detect the `redirectTo` prop and then compute this variable in computed properties, and rename to `isRedirectToPlio`

pritamps:
> * Sometimes I added multi-line comments to communicate the logic. Once we have a documentation process ready, I think we can transfer these comments and reduce them to single-line.

This is good! There is a purpose for comments in code, and a different purpose for documentation outside it :)

package-lock.json
```diff
@@ -0,0 +1,17609 @@
+{
```
deepansh96: we should not commit package-lock.json!
put it in .gitignore and don't commit it
Otherwise if it gets commited once it'll always show up as code change in PRs
I dont think there is a .gitignore in this PR?

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+      <input
+        v-model="singleUserID"
+        type="text"
+        placeholder="Student SRN"
+        required
+        @keypress="isValidSRNFormat($event)"
+        :class="inputStyleClass"
+      />
+      <button @click="this.processForm" :class="buttonStyleClass">SUBMIT</button>
+    </div>
```
deepansh96: I feel this is unnecessary
You can use the piece of code that's written below only
The only addition will be that this `v-if="isRedirectToPlio"` is applied on the addField and removeField buttons so one cannot do that if it's a plio being redirected to

See my comment below

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+          v-model="input.userID"
+          type="text"
+          placeholder="Student SRN"
+          required
+          @keypress="isValidSRNFormat($event)"
+        />
+        <inline-svg
+          @click="addField(index, userIDList)"
+          :src="require('@/assets/images/add-button.svg')"
+        ></inline-svg>
```
```suggestion
        <inline-svg
          v-if="!isRedirectToPlio"
          @click="addField(index, userIDList)"
          :src="require('@/assets/images/add-button.svg')"
        ></inline-svg>
```

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+        <inline-svg
+          v-show="isAnyUserIDPresent"
+          @click="removeField(index, userIDList)"
+          class="ml-2 cursor-pointer"
+          :src="require('@/assets/images/delete-button.svg')"
+        ></inline-svg>
```
```suggestion
        <inline-svg
          v-show="isAnyUserIDPresent && !isRedirectToPlio"
          @click="removeField(index, userIDList)"
          class="ml-2 cursor-pointer"
          :src="require('@/assets/images/delete-button.svg')"
        ></inline-svg>
```

src/components/SRNEntry.vue
deepansh96:
```suggestion
    isRedirectToPlio: Boolean,
```

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+    redirectTo: String,
+    redirectID: String,
+    purpose: String,
+    subPurpose: String,
+    plioUse: Boolean,
+  },
+  data() {
+    return {
+      userIDList: [{ userID: "" }],
+      singleUserID: "", //is a temp hack
```
deepansh96: Will not be needing this if we use userIDList directly

src/components/SRNEntry.vue
deepansh96: Yes we should do that!

src/components/SRNEntry.vue
deepansh96: You mean the reverse of it? like `isUserIDListEmpty`?

src/pages/Homepage.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div>
+    <SRNEntry
+      :redirectTo="redirectTo"
+      :redirectID="redirectID"
+      :purpose="purpose"
+      :subPurpose="subPurpose"
+      :plioUse="plioUse"
```
deepansh96: No need to pass this as a prop

You're already passing `redirectTo` and you can compute `isRedirectToPlio` using `redirectTo` inside that component itself

src/pages/Homepage.vue
```diff
@@ -0,0 +1,66 @@
...
+    />
+    <div class="mx-auto" v-if="!plioUse">
```
deepansh96:
```suggestion
    <div class="mx-auto" v-if="!isRedirectToPlio">
```

src/pages/Homepage.vue
deepansh96: Yes, I would say keep the variable, rename it to `isRedirectToPlio` and actually compute that from `redirectTo`

src/pages/Homepage.vue
deepansh96: Pritam meant that plio here can only be one example so you should mention that in the comments as an example only. Anyway, see my comments for the props

src/pages/Homepage.vue
deepansh96: or `purposeParams`?
We might need `params` for something else as well in the future maybe!

src/services/firebase.py
```diff
@@ -0,0 +1,11 @@
+import firebase_admin
```
deepansh96: What is a python file doing here :p

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+    },
+    inputStyleClass() {
+      return "flex border py-2 mx-auto px-3 text-grey-darkest";
+    },
+    isAnyUserIDPresent() {
+      return this.userIDList.length > 1;
+    },
+  },
+  methods: {
+    //checking to see if each char typed by user is only a letter or number
```
deepansh96: put this comment inside the method!

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+      if (/^[a-z0-9]+$/.test(char)) return true;
+      else e.preventDefault();
+    },
+    addField(value, list) {
+      list.push({ value: "" });
+    },
+    removeField(index, list) {
+      list.splice(index, 1);
+    },
+    //this method constructs the URL based on the redirectTo param
```
deepansh96: inside the method

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+    processForm() {
+      if (this.redirectTo == "plio") {
```
deepansh96: Can directly use `isRedirectToPlio` here where this `isRedirectToPlio` is a compute variable like this
```js
isRedirectToPlio() {
}
```

deepansh96 (changes requested):

src/components/SRNEntry.vue
sudheshna-donthineni: The problem with that is, `userIDList ` is an object, whereas `singleUserID` is a string. Player is expecting the `unique_id` to be a string, so when I send `userIDList`, it does not detect it.

src/assets/images/add-button.svg
```diff
@@ -0,0 +1,5 @@
+<svg viewBox="0 0 24 24" width="24" height="24">
```
pritamps, line 1: Sab cheez ke liye SVG? We should see if it's possible with pure CSS. Will probably be possible! Not for this PR though.

SVGs are in general more computationally expensive

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
+<template>
+  <section>
+    <div v-if="plioUse" class="flex flex-col mt-8">
```
pritamps: Not `plioUse`. This component ka code shouldn't know about use case. this `v-if` should just be `ifSingleSRNonly` or something like that

src/components/SRNEntry.vue
pritamps: Haan whatever way. `length` is unnecessary operation. micro optimzation :D

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+    isAnyUserIDPresent() {
+      return this.userIDList.length > 1;
+    },
+  },
+  methods: {
+    //checking to see if each char typed by user is only a letter or number
+    isValidSRNFormat(e) {
+      let char = String.fromCharCode(e.keyCode);
+      if (/^[a-z0-9]+$/.test(char)) return true;
+      else e.preventDefault();
```
pritamps, line 79: Maybe, just to add some  extra checking, return false if the string matches:

```
0000000...
1111...
222...
...
999....
```

Unless any of these are alreayd in the list. We can also keep a global variable for the list

src/components/SRNEntry.vue
```diff
@@ -0,0 +1,105 @@
...
+      list.splice(index, 1);
+    },
+    //this method constructs the URL based on the redirectTo param
+    processForm() {
+      if (this.redirectTo == "plio") {
+        const redirectURL = process.env.VUE_APP_BASE_URL_PLIO_STAGING;
+        let url = new URL(redirectURL + this.redirectID); //adds plioID to the base plio link
+        //adds params; api key and student SRN
+        let queryparams = new URLSearchParams({
+          api_key: process.env.VUE_APP_AF_STAGING_API_KEY,
```
pritamps: STAGING and PROD both should be here depending on the environment, no?

src/pages/Homepage.vue
pritamps: `plioUse` should be resolved to `isSingleEntryOnly` or something like that

src/pages/Homepage.vue
pritamps: These comments should deal with the auth layer. "The plio link" can be replaced with "The link is built based on the `subPurpose` field, for example

pritamps (changes requested):
Almost there, I think! Cool! We should talk a bit about architecture after shipping, and before you continue implementing

src/assets/images/add-button.svg
sudheshna-donthineni: I'll keep this in mind next time.

src/components/SRNEntry.vue
sudheshna-donthineni: I refactored it like Deepansh said. Used a computed property `isRedirectToPlio`. It computes the value based on the `redirectTo` param.

src/components/SRNEntry.vue
sudheshna-donthineni: Is this required for now? Because the task that I had paused before this, was getting data from firebase. So I will be validating the SRN against that dataset anyway.

deepansh96: Not super necessary for the monday's pilot, but if it's gonna take some 20-30 minutes to add and test this, then can do!

src/components/SRNEntry.vue
deepansh96: Configured this

src/components/SRNEntry.vue
deepansh96: Let's do it after the monday pilot

PR #6, opened 2021-08-11
Fix for issue#4

### LINKED ISSUE #4: On a mobile phone - can type alphabets in the SRN box (should only be numbers)
Also, in mobile phone, entering something like `lajksdalns`(10 alphabets) in the SRN box keeps the button inactive but entering something like `asdqwe12345` (10 characters, but at least one is letter) makes the button active, which should not be the case

sudheshna-donthineni:
**Problem:** On mobile, the user could type alphabets as well.
**Fix:** Added slice function to the input variable once its length crosses 10. 
Also, added `inputmode=numeric` to force numeric keypad on Andriod phones.

Sidenote: Also, removed the usage of SVG and used plain CSS to create add and minus buttons. Won't be a need for the plio use case though.


src/components/SRNEntry.vue
```diff
@@ -65,19 +64,21 @@ export default {
...
+      return !this.isSingleEntryOnly && this.userIDList.length > 1;
     },
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
     isSubmitButtonDisabled() {
       return !this.isAnyUserIDPresent || this.invalidInputMessage != "";
     },
+    isUserValidated() {
+      return !this.isSingleEntryOnly && this.userIDList[0]["userID"].length == 10;
```
deepansh96: Use maxLength here

PR #8, opened 2021-08-31
Advanced Validation of user

sudheshna-donthineni:
- Once the user clicks submit, an API call is made. `firebaseAPI.checkUserExists`
- If the user is authenticated, then they are directed to the Plio.
- If the user is not authenticated, then the API call is made again and checked again.
- At this step, whether the user is authenticated or not, they are directed to the Plio.

- The variable `validateCount` is kept track of to see how many times a user is authenticated. 
- While sending an SQS message, this variable tells us whether to set the flag to valid or not.
- The variable has three states:
0 -> this means the user is authenticated on the first try and the flag in SQS is set to valid.
1 -> this means the user wasn't authenticated on the first try and is sent for re-authenticated.  This state helps with displaying the error message. If the user is authenticated now, then the flag in SQS to set to valid. If not, then the flag is set to invalid and the state is changed to 1.
2 -> this means the user hasn't been authenticated.


pritamps:
> * Once the user clicks submit, an API call is made. `firebaseAPI.checkUserExists`
> * If the user is authenticated, then they are directed to the Plio.
> * If the user is not authenticated, then the API call is made again and checked again.
> * At this step, whether the user is authenticated or not, they are directed to the Plio.
> * The variable `validateCount` is kept track of to see how many times a user is authenticated.
> * While sending an SQS message, this variable tells us whether to set the flag to valid or not.
> * The variable has three states:
>   -1 -> this means the user is authenticated on the first try and the flag in SQS is set to valid.
>   0 -> this means the user wasn't authenticated on the first try and is sent for re-authenticated.  This state helps with displaying the error message. If the user is authenticated now, then the flag in SQS to set to valid. If not, then the flag is set to invalid and the state is changed to 1.
>   1 -> this means the user hasn't been authenticated.

-1 for authenticated and 1 for not authenticated? Seems a bit weird. Would suggest the following:

0 - not authenticated
1 - authenticated
2 - Invalid
-1 - Unknown status or some other failure

src/components/SRNEntry.vue
```diff
@@ -55,7 +59,16 @@ export default {
     return {
       userIDList: [{ userID: "" }],
       invalidInputMessage: null,
+      validated: true,
       maxLength: 10,
+      /*this count is used to check how many times the user has been validated.
```
pritamps: Don't use counts. Use an enum with string constants. Will make the code much more readable

src/services/API/checkUser.js
```diff
@@ -0,0 +1,27 @@
+import axios from 'axios';
```
pritamps: @deepansh96 can you guide here and help @sudheshna-donthineni extract all network calls to a service?

pritamps (changes requested):

src/components/SRNEntry.vue
sudheshna-donthineni: Using an enum `loginState : {
        UNAUTHENTICATED : 0,
        AUTHENTICATED : 1,
        INVALID : 2
      },`
      
  but I'll still need to maintain a global variable to compare with the enum, right?

src/components/SRNEntry.vue
```diff
@@ -55,7 +59,16 @@ export default {
     return {
       userIDList: [{ userID: "" }],
       invalidInputMessage: null,
+      validated: true,
```
```suggestion
      doesUserExists: true,
```
Suggestion - 
Two things, for boolean values, it's super helpful to have `is` or `does` in the beginning.
And I think this variable name accurately represents the output of `checkUserExists` method

src/components/SRNEntry.vue
```diff
@@ -55,7 +59,16 @@ export default {
...
       maxLength: 10,
```
```suggestion
      maxLengthOfSRN: 10,
```

src/components/SRNEntry.vue
deepansh96: Or maybe, we can keep the same variable but treat it as an actual count. Like, the first value will be 0. That means the user has been validated 0 times. And over time as the user gets validated, this could will increase from 0 to 1 and then from 1 to 0. `-1` was not super clear to me

src/components/SRNEntry.vue
```diff
@@ -82,6 +95,9 @@ export default {
         this.userIDList[0]["userID"].length > this.maxLength - 1
       );
     },
+    isincorrectLogin() {
+      return !this.validated && this.validateCount == 0;
+    },
```
deepansh96: this is very specific. `isIncorrectLogin` can mean both things, incorrect login once or twice. 
As you're using this just in the HTML v-if part, I would suggest using it like this
`v-if="!validated && validateCount == 0"` and getting rid of the computed variable

src/components/SRNEntry.vue
```diff
@@ -31,9 +31,12 @@
       <span class="errorStyleClass" v-if="invalidInputMessage">{{
         invalidInputMessage
       }}</span>
-      <span class="errorStyleClass" v-if="isincorrectLogin">{{
-        invalidLoginMessage
-      }}</span>
+      <span
+        class="errorStyleClass"
+        v-if="!this.doesUserExist && this.validateCount == 1"
```
deepansh96: No need to use `this` keyword when writing JS in HTML template

src/components/SRNEntry.vue
```diff
@@ -59,15 +62,10 @@ export default {
...
-      validated: true,
-      maxLength: 10,
-      /*this count is used to check how many times the user has been validated. 
-      For now, the count has three possible states : -1 -> the initial state, if the user is validated in the first try. This will set the flag to valid
-      0 -> user has been authenticated once, and is being authenticated again. If yes, then the flag is set to valid. 
-      Otherwise, the state changes to 1, this means the user has tried getting authenticated twice, but is invalid. 
-
-      */
-      validateCount: -1,
+      doesUserExist: false,
```
deepansh96: Add one comment like this
```js
doesUserExist: false, // whether the user exists in the backend database
```

deepansh96: What I do personally, before I make the PR, I add one liner comments to everything, methods, computed variables, data properties. Helps me when I revisit the code after months

src/services/validation.js
```diff
@@ -0,0 +1,22 @@
+import firebaseAPI from "@/services/API/checkUser.js";
+import { sendPlio } from "./sendPlio";
```
deepansh96: Use absolute path using "@" here

src/services/validation.js
```diff
@@ -0,0 +1,22 @@
...
+
+export async function validateUser(userID, validateCount, isSingleEntryOnly, redirectID, doesUserExist){
```
deepansh96: The name should be something like `validateSRNUser` no?
Because slowly we'll add more functions in this file for different kinds of validation right?

src/services/validation.js
```diff
@@ -0,0 +1,22 @@
...
+    //response tells us if the user is authenticated.
+    if(!(/^([0-9])\1*$/.test(userID)) && userID.toString()[0] == '1'){
+        doesUserExist = await firebaseAPI.checkUserExists(userID);
+    }
+    // this condition checks if the user is getting authenticated the first time. Just shows an error message.
+    if (!doesUserExist && validateCount == 0) {
+    validateCount = 1;
+    var invalidLoginMessage = "Please enter correct SRN / कृपया सही SRN दर्ज करें";
```
deepansh96: because this variable is declared in the if block, if the compiler doesn't reach this if condition, this will be undefined when you're returning it at the end. 
You can declare the variable outside the if block, that way it will be accessible everywhere

src/services/validation.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+    //this condition checks the second time, since still not valid, just changes the flag and continues with the plio.
+    else if (!doesUserExist && validateCount == 1) {
+    validateCount = 2;
+    sendPlio(isSingleEntryOnly, userID, redirectID);
+    } else {
+    sendPlio(isSingleEntryOnly, userID, redirectID);
+    }
+    return [doesUserExist, validateCount, invalidLoginMessage]
```
deepansh96: When you want to return multiple things from a function, best to wrap them up in an object instead of an array.
```js
return {
    doesUserExist: doesUserExist,
    validateCount: validateCount,
    invalidLoginMessage: invalidLoginMessage
}
```
You'll have to make changes accordingly where you're reading this array

src/components/SRNEntry.vue
```diff
@@ -143,20 +124,19 @@ export default {
       const userID = parseInt(this.userIDList["0"]["userID"]);
 
       //response tells us if the user is authenticated.
-      this.validated = await firebaseAPI.checkUserExists(userID);
+      const userIsValidated = validateUser(
```
deepansh96: It's not a good practice to use `const` for objects.
Here, `userIsValidated` is an object (yes, arrays are objects too).
`const` variables cannot be unassigned but they can be mutated for example `userIsValidated[2] = "new value"`
This statement is valid. So there's no point to use `const`
You should use simply `let`

deepansh96 (changes requested):

src/services/validation.js
sudheshna-donthineni: Yeah I realized that and changed it in a later commit. So if you see, it is `validateSRN`

src/services/validation.js
sudheshna-donthineni: even if it is in the same directory?

deepansh96: Yep
Just to stay consistent
In case we change the directories in the future, it'll be helpful to know every import part is relative to the root src directory

PR #9, opened 2021-09-05
Deleted unused packages

sudheshna-donthineni:


src/store/index.js
```diff
@@ -1,25 +1,8 @@
 import { createStore } from "vuex";
-import createPersistedState from "vuex-persistedstate";
-import SecureLS from "secure-ls";
 import auth from "./modules/auth";
 
-var localStorage = new SecureLS({ isCompression: false });
-
 export default createStore({
```
pritamps, line 4: This change should not be in this PR?

sudheshna-donthineni: We didn't want the secure-ls package anymore, so I deleted the code that uses it.

PR #11, opened 2021-09-13
Added AWS ENV variables to github actions

sudheshna-donthineni:


.github/workflows/deploy_to_s3_staging.yml
```diff
@@ -5,7 +5,7 @@ name: Deploy to staging
 on:
   pull_request:
   push:
-    branches: ["master"]
+    branches: ["main"]
```
pritamps, line 8: Why this change? We're using `main` for this repo as default?

pritamps: Also, why not this change for `prod`? Is it part of the config?

pritamps (approved):
Approved -- just a small question

pritamps:
Looks like the GH action failed also?

PR #12, opened 2021-09-17
Live class use case

sudheshna-donthineni:
This code abstracts filtering out destination, which then leads to the respective functions.


src/services/API/redirectToDest.js
```diff
@@ -0,0 +1,24 @@
+export function redirectToPlio(userID, redirectID) {
+    //this method constructs the URL based on the redirectTo param
```
pritamps: Rename the file `redirectToDestination`. No need for `dest`.

src/services/filterDest.js
```diff
@@ -0,0 +1,13 @@
+import { redirectToPlio, redirectToMeet } from "@/services/API/redirectToDest.js";
+
+export function filterDestination(redirectTo, purposeParams, userID, redirectID){
+    if(purposeParams == "plio"){
```
pritamps: Use a switch-case here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

pritamps (changes requested):
Looks good! Just a small change and you can merge!

pritamps (commented):
Why were the `yml` files deleted?

src/services/filterDest.js
```diff
@@ -0,0 +1,13 @@
+import { redirectToPlio, redirectToMeet } from "@/services/API/redirectToDest.js";
+
```
pritamps: Same here `filterDestination` for the filename

PR #17, opened 2021-10-06
Multiple SRN feature

sudheshna-donthineni:
- Added support to have more than one SRN entry. 
- Split the process method `processForm` into `authenticateSRN` and `processForm`.
- Also, made the change to the input array. The `userIDList` also contains a boolean variable for each SRN which tells whether the SRN is valid or not. This helps to create separate rows in Bigquery for each SRN. 
- Have removed the userDataValidated flag in SQS message as the userIDList array itself contains the userID and the respective valid flag.

Example of the SQS message: `{'dateTime': '2021-10-06T10:17:51.887Z', 'purpose': {'type': 'attendance', 'subType': 'liveclass', 'params': {'platform': 'meet', 'id': 'eig-ioxa-rvj'}}, 'authType': 'SRN', 'user': {'values': [{'userID': '1111111111', 'valid': True}, {'userID': '1231231231', 'valid': False}, {'userID': '1111111110', 'valid': True}]}}`

### DIFF
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -1,12 +1,29 @@
 <template>
-  <section>
-    <div>
-      <label>Enter your SRN / अपना SRN दर्ज करें</label>
-      <div
-        class="multipleStudentStyle"
-        v-for="(input, index) in userIDList"
-        :key="`IDInput-${index}`"
-      >
+  <!-- loading spinner -->
+  <div v-if="isLoading" class="h-full w-full fixed z-50">
+    <div class="flex mx-auto w-full h-full">
+      <span class="material-icons text-black text-4xl m-auto animate-spin">
+        autorenew
+      </span>
+    </div>
+  </div>
+  <!-- main div -->
+  <div
+    class="flex flex-col my-auto h-full py-32 space-y-6"
+    :class="{ 'opacity-20 pointer-events-none': isLoading }"
+  >
+    <!-- title -->
+    <p class="text-xl sm:text-2xl md:text-2xl lg:text-3xl xl:text-4xl mx-auto font-bold">
+      Enter your SRN / अपना SRN दर्ज करें
+    </p>
+    <!-- input options and delete options icon -->
+    <div
+      class="flex flex-row justify-center"
+      v-for="(input, index) in userIDList"
+      :key="`IDInput-${index}`"
+      :class="{ 'pl-12': ifUserEnteredMoreThanOne }"
+    >
+      <div>
         <input
           v-model="input.userID"
           type="tel"
@@ -15,44 +32,57 @@
           placeholder="Your SRN / आपका SRN"
           required
           @keypress="isValidNumericEntry($event)"
-          class="inputStyleClass"
+          class="border-2 rounded-sm p-4 mx-auto border-gray-500 focus:border-gray-800 focus:outline-none"
+          :class="calculateInputboxStyleClasses(index)"
           @input="updateValue($event, index)"
         />
+      </div>
 
+      <div class="my-auto px-3" v-show="ifUserEnteredMoreThanOne">
         <button
-          class="material-icons minusButton"
-          v-show="ifUserEnteredMoreThanOne"
+          class="material-icons text-red-500"
           @click="removeField(index, userIDList)"
         >
           remove_circle
         </button>
       </div>
+    </div>
 
-      <span class="errorStyleClass" v-if="invalidInputMessage">{{
-        invalidInputMessage
-      }}</span>
-      <span class="errorStyleClass" v-if="!isUserValid && validateCount == 1">{{
-        invalidLoginMessage
-      }}</span>
-      <div v-if="isLoading" class="loadingSpinner"></div>
-      <div class="flex flex-row my-auto multiple-div" v-if="isAddButtonAllowed">
-        <button @click="addField" class="addButtonStyleClass">
-          <span class="material-icons addButton"> add_box </span>
-          <p>
+    <!-- invalid input and login message  -->
+    <span class="mx-auto text-red-700 text-base mb-1" v-if="invalidInputMessage">{{
+      invalidInputMessage
+    }}</span>
+    <span
+      class="mx-auto text-red-700 text-base mb-1"
+      v-if="!isUserValid && validateCount == 1"
+      >{{ invalidLoginMessage }}</span
+    >
+    <!-- add srn button -->
+    <div class="my-auto" v-if="isAddButtonAllowed">
+      <button
+        @click="addField"
+        class="flex flex-row mx-auto p-2 items-center border-2 rounded-xl bg-gray-200 btn"
+      >
+        <span class="material-icons text-green-500 text-4xl pr-2">
+          add_circle_outline
+        </span>
+        <div class="border-l-2 border-gray-500 pl-4">
+          <p class="leading-tight">
             Add another SRN <br />
             एक और SRN दर्ज करें
           </p>
-        </button>
-      </div>
-      <button
-        @click="processForm"
-        class="buttonStyleClass"
-        :disabled="isSubmitButtonDisabled"
-      >
-        SUBMIT / जमा करें
+        </div>
       </button>
     </div>
-  </section>
+    <!-- submit button -->
+    <button
+      @click="processForm"
+      class="bg-primary hover:bg-primary-hover text-white font-bold shadow-xl uppercase text-lg mx-auto p-4 mt-4 rounded disabled:opacity-50 btn"
+      :disabled="isSubmitButtonDisabled"
+    >
+      SUBMIT / जमा करें
+    </button>
+  </div>
 </template>
 
 <script>
@@ -82,17 +112,20 @@ export default {
     };
   },
   computed: {
+    userIdListLength() {
+      return this.userIDList.length;
+    },
     isAnyUserIDPresent() {
       //checks if any userID has been typed
       return (
         this.userIDList != undefined &&
-        this.userIDList.length > 0 &&
+        this.userIdListLength > 0 &&
         this.userIDList[0]["userID"] != ""
       );
     },
     ifUserEnteredMoreThanOne() {
       //used when multiple SRN's will be allowed to typed
-      return !this.isSingleEntryOnly && this.userIDList.length > 1;
+      return !this.isSingleEntryOnly && this.userIdListLength > 1;
     },
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
@@ -109,22 +142,27 @@ export default {
       //if multiple SRN's are allowed, then this helps with the activation of the + button, to add more SRN's
       return (
         !this.isSingleEntryOnly &&
-        this.isInputLengthComplete &&
-        this.userIDList.length < numberOfSRNsAllowed
+        !this.isCurrentEntryIncomplete &&
+        this.userIdListLength < numberOfSRNsAllowed
       );
     },
-    isInputLengthComplete() {
-      return this.getLatestEntry["userID"].length == this.maxLengthOfSRN;
-    },
-
     isCurrentEntryIncomplete() {
-      return this.getLatestEntry["userID"].length < this.maxLengthOfSRN - 1;
+      return this.getLatestEntry["userID"].length < this.maxLengthOfSRN;
     },
     getLatestEntry() {
       return this.userIDList.slice(-1)[0];
     },
   },
   methods: {
+    calculateInputboxStyleClasses(index) {
+      return [
+        {
+          "border-red-600 focus:border-red-600":
+            this.invalidInputMessage && index == this.userIdListLength - 1,
+          "pointer-events-none opacity-30": index < this.userIdListLength - 1,
+        },
+      ];
+    },
     isValidNumericEntry(e) {
       //checking to see if each char typed by user is only a number
       if (e.keyCode >= 48 && e.keyCode <= 57) return true;
@@ -148,7 +186,6 @@ export default {
     setValidFlag() {
       this.getLatestEntry["valid"] = this.isUserValid;
     },
-
     async addField() {
       //for adding another field, the previously entered ID is validated against the database
       const latestUserID = parseInt(this.getLatestEntry["userID"]);
@@ -177,7 +214,9 @@ export default {
     },
     updateValue(event, index) {
       //checks if the 10 characters are entered
-      if (event.target.value.length < this.maxLengthOfSRN) {
+      if (event.target.value.length == 0) {
+        this.invalidInputMessage = "";
+      } else if (event.target.value.length < this.maxLengthOfSRN) {
         this.invalidInputMessage = "Please type 10 numbers / कृपया १० संख्या टाइप करें";
         this.resetInvalidLoginMessage();
       } else {
@@ -249,34 +288,10 @@ export default {
 </script>
 
 <style lang="postcss">
-div {
-  @apply flex flex-col mt-8;
+.btn {
+  border-bottom: outset;
 }
-label {
-  @apply mb-2 mx-auto uppercase font-bold text-lg;
-}
-.inputStyleClass {
-  @apply flex border py-2 mx-auto px-3 h-12;
-}
-.buttonStyleClass {
-  @apply bg-primary hover:bg-primary-hover text-white uppercase text-lg mx-auto p-4 mt-4 rounded disabled:opacity-50;
-}
-.addButtonStyleClass {
-  @apply bg-white flex flex-row mx-auto p-2 items-center space-x-12;
-}
-.errorStyleClass {
-  @apply mx-auto text-red-700 text-base mb-1;
-}
-.multipleStudentStyle {
-  @apply flex flex-row w-1/2 md:w-1/6 lg:w-1/6  mx-auto items-center;
-}
-.addButton {
-  @apply text-green-500;
-}
-.minusButton {
-  @apply text-red-500;
-}
-.loadingSpinner {
-  @apply mx-auto w-4 h-4 border-4 border-primary border-dotted rounded-full animate-spin;
+.btn:active {
+  border-bottom: unset;
 }
 </style>
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -34,17 +34,12 @@
         invalidLoginMessage
       }}</span>
 
-<<<<<<< HEAD
+
       <div class="flex flex-row my-auto multiple-div" v-if="isAddButtonAllowed">
         <button @click="addField" class="addButtonStyleClass">
           <span class="material-icons addButton"> add_box </span>
           Add another SRN <br />
           एक और SRN दर्ज करें
-=======
-      <div class="flex flex-row my-auto multiple-div" v-if="isUserValidated">
-        <button @click="addField(userIDList)" class="addButtonStyleClass">
-          Add another SRN / एक और SRN दर्ज करें
->>>>>>> e3c51f09b5d9b31d8178a09c5d9ad3a051a12072
         </button>
       </div>
       <button
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -18,13 +18,14 @@
           class="inputStyleClass"
           @input="updateValue($event, index)"
         />
-        <div
-          class="minus-sign"
+
+        <button
+          class="material-icons minusButton"
           v-show="ifUserEnteredMoreThanOne"
           @click="removeField(index, userIDList)"
         >
-          <span class="material-icons minusButton"> remove_circle </span>
-        </div>
+          remove_circle
+        </button>
       </div>
 
       <span class="errorStyleClass" v-if="invalidInputMessage">{{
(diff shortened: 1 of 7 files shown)


src/components/SRNEntry.vue
```diff
@@ -9,24 +9,35 @@
...
+      </div>
+
       <span class="errorStyleClass" v-if="invalidInputMessage">{{
         invalidInputMessage
       }}</span>
       <span class="errorStyleClass" v-if="!isUserValid && validateCount == 1">{{
         invalidLoginMessage
       }}</span>
 
+      <div class="flex flex-row my-auto multiple-div" v-if="isUserValidated">
```
pritamps: This should check if all users before are validated right? Or, if you only let them add another user after the first one, that the last one is validated.

Another thing you should do is to make the valid users uneditable. Once added, is done

src/components/SRNEntry.vue
```diff
@@ -52,8 +63,8 @@ export default {
   },
   data() {
     return {
-      userIDList: [{ userID: "" }],
-      invalidInputMessage: null,
+      userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
```
pritamps, line 105: Ah this is a good idea

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
+        this.userIDList.slice(-1)[0]["userID"].length < this.maxLengthOfSRN - 1
+      );
     },
     isUserValidated() {
       //if multiple SRN's are allowed, then this helps with the activation of the + button, to add more SRN's
       return (
         !this.isSingleEntryOnly &&
-        this.userIDList[0]["userID"].length > this.maxLengthOfSRN - 1
+        this.userIDList.slice(-1)[0]["userID"].length > this.maxLengthOfSRN - 1 &&
+        this.userIDList.length < 10
```
pritamps: Aren't all SRNs the same lengh? Why this check of `>` and `<`

src/components/SRNEntry.vue
pritamps: Maybe this is what is being done already looking at `isUserValidated`? Is this deployed on staging?

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
       );
     },
   },
   methods: {
     isValidSRNFormat(e) {
       //checking to see if each char typed by user is only a number
-      let char = String.fromCharCode(e.keyCode);
-      if (/^[0-9]+$/.test(char)) return true;
+      if (e.keyCode >= 48 && e.keyCode <= 57) return true;
```
pritamps, line 168: Oooh ASCII :)

pritamps (commented):
This looks good! Just some comments. Is this deployed on staging already?

src/components/SRNEntry.vue
sudheshna-donthineni: The first statement checks the length of the individual SRN, whereas the second statement caps the number of SRNs you can add. For now, the ops team decided to allow 10 SRNs at a time.

src/components/SRNEntry.vue
sudheshna-donthineni: Yeah, that is happening, but it's happening in the `addField` method. SRN gets authenticated there.

pritamps: `addField` shouldn't authenticate, no?

src/components/SRNEntry.vue
pritamps: SRNs are not the same length? Why not use `==` rather than `>`?

src/components/SRNEntry.vue
sudheshna-donthineni: I didn't say that correctly. There is a method `authenticateSRN`. `addField` calls this method as it is a pre-requisite for adding another SRN.

src/components/SRNEntry.vue
```diff
@@ -9,24 +9,35 @@
...
           @keypress="isValidSRNFormat($event)"
           class="inputStyleClass"
-          @input="updateValue"
+          @input="updateValue($event, index)"
         />
-       </div>
-       
+        <div
+          class="minus-sign"
+          v-show="ifUserEnteredMoreThanOne"
```
deepansh96: Why not `v-if`?

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
       return this.redirectTo == "plio";
     },
     isSubmitButtonDisabled() {
-      return !this.isAnyUserIDPresent || this.invalidInputMessage != "";
+      //submit button is disabled if no SRN has been typed or input is invalid or SRN hasn't been completely typed
+      return (
+        !this.isAnyUserIDPresent ||
+        this.invalidInputMessage != "" ||
+        this.userIDList.slice(-1)[0]["userID"].length < this.maxLengthOfSRN - 1
+      );
```
deepansh96: In this piece of code `this.userIDList.slice(-1)[0]["userID"].length < this.maxLengthOfSRN - 1`, why are we only checking the latest SRN entry? Is it because they will only be able to add another SRN entry if all others are valid inputs right?
Also, why are we saying that if the length < (10 - 1), that is, length <= 8, then submit button should be disabled.
Does that mean that 9 digit SRN numbers are allowed?

Also, I think you should make a computed property for this -- something like, 
```js
isCurrentEntryIncomplete() {
  return this.userIDList.slice(-1)[0]["userID"].length < this.maxLengthOfSRN - 1
}
```
and then use this computed value in your method

Also, I think you should also change the invalidInputMessage and invalidLoginMessage into computed properties, it's hard to explain right now, I'll make a separate PR proposing my changes

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
     },
     isUserValidated() {
       //if multiple SRN's are allowed, then this helps with the activation of the + button, to add more SRN's
       return (
         !this.isSingleEntryOnly &&
-        this.userIDList[0]["userID"].length > this.maxLengthOfSRN - 1
+        this.userIDList.slice(-1)[0]["userID"].length > this.maxLengthOfSRN - 1 &&
```
deepansh96: This line you have used already before once, so it would be super helpful to make it into a computed variable

src/components/SRNEntry.vue
deepansh96: I agree. Both `==` and `>` or `<` works logically, but while reading the code, `==` makes the most sense

deepansh96: Also, the maximum allowed entries, make it a global const instead of hardcoding a value `10` here

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
     isSubmitButtonDisabled() {
-      return !this.isAnyUserIDPresent || this.invalidInputMessage != "";
+      //submit button is disabled if no SRN has been typed or input is invalid or SRN hasn't been completely typed
+      return (
+        !this.isAnyUserIDPresent ||
+        this.invalidInputMessage != "" ||
+        this.userIDList.slice(-1)[0]["userID"].length < this.maxLengthOfSRN - 1
+      );
     },
     isUserValidated() {
```
deepansh96: The name is `isUserValidated` but in this computed property, we're not doing advanced validation right?

deepansh96: So you just show a plus icon even if the SRN is not a real SRN right?

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
       return (
         !this.isSingleEntryOnly &&
-        this.userIDList[0]["userID"].length > this.maxLengthOfSRN - 1
+        this.userIDList.slice(-1)[0]["userID"].length > this.maxLengthOfSRN - 1 &&
+        this.userIDList.length < 10
       );
     },
   },
   methods: {
     isValidSRNFormat(e) {
```
```suggestion
    allowNumericEntriesOnly(e) {
```
As this is a method, the naming should be such that it tells us what the function does.

src/components/SRNEntry.vue
```diff
@@ -9,24 +9,35 @@
...
       <span class="errorStyleClass" v-if="!isUserValid && validateCount == 1">{{
         invalidLoginMessage
       }}</span>
 
+      <div class="flex flex-row my-auto multiple-div" v-if="isUserValidated">
+        <button @click="addField(userIDList)" class="addButtonStyleClass">
+          <div class="plus-sign mr-3" @click="addField(index, userIDList)"></div>
+          Add another SRN / एक और SRN दर्ज करें
+        </button>
+      </div>
```
deepansh96: Why is there an @click listener added to a div element here?
The one attached to the button should be enough right?
It looks like the div is not being used anyhow

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
-      let char = String.fromCharCode(e.keyCode);
-      if (/^[0-9]+$/.test(char)) return true;
+      if (e.keyCode >= 48 && e.keyCode <= 57) return true;
       else e.preventDefault();
     },
-    addField(value, list) {
-      list.push({ value: "" });
+    async addField(list) {
+      //for adding another field, the previously entered ID is validated against the database
+      const userID = parseInt(this.userIDList.slice(-1)[0]["userID"]);
```
deepansh96: - How about making this a computed variable? `latestEntry`? That way you can just use it like
`const userID = parseInt(this.latestEntry)`

- also, 
```suggestion
      const latestUserID = parseInt(this.userIDList.slice(-1)[0]["userID"]);
```

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
+      if (!isNaN(userID)) {
+        await this.authenticateSRN(userID);
```
deepansh96: It's about time we added a loading spinner!
We need to show loading to the user while they wait for authentication to happen

src/components/SRNEntry.vue
deepansh96: Also, in one method call you're just using `userIdList` and in another one you're using two parameters, `index` and `userIDlist`?

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
     isValidSRNFormat(e) {
       //checking to see if each char typed by user is only a number
-      let char = String.fromCharCode(e.keyCode);
-      if (/^[0-9]+$/.test(char)) return true;
+      if (e.keyCode >= 48 && e.keyCode <= 57) return true;
       else e.preventDefault();
     },
-    addField(value, list) {
-      list.push({ value: "" });
+    async addField(list) {
```
deepansh96: Why is there a need for passing the list here as a parameter?
You can just use `this.userIDList` right?

src/components/SRNEntry.vue
deepansh96: Because of the method call with 2 parameters, whenever I press plus icon I get an error in console because you're passing `index` but the method is expecting a `userIDList`

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
+      //for adding another field, the previously entered ID is validated against the database
+      const userID = parseInt(this.userIDList.slice(-1)[0]["userID"]);
+      if (!isNaN(userID)) {
+        await this.authenticateSRN(userID);
+        //only if the SRN is valid or if the user is entering a SRN for the second time,the loop is entered
+        if (this.isUserValid || this.validateCount > 1) {
+          //setting the flag of the SRN
+          this.userIDList.slice(-1)[0]["valid"] = this.isUserValid;
+          list.push({ userID: "", valid: false });
```
deepansh96: You can make this a method
```js
methods: {
  addNewEmptyField() {
      this.userIDList.push({ userID: "", valid: false})
  }
}
```
Makes the code more readable and if in future we change the format of our `userIDList`, we won't need to change everywhere

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
+          //resetting flags to default for processing next SRN
+          this.isUserValid = false;
+          this.validateCount = 0;
+        }
+      }
     },
     removeField(index, list) {
+      //resetting all messages to default before deleting the input field
+      this.invalidInputMessage = "";
+      this.invalidLoginMessage = "";
```
deepansh96: Again, can make methods for this
```js
methods: {
	resetInvalidInputMessage() {
		this.invalidInputMessage = ""
	},
	resetInvalidLoginMessage() {
		this.invalidLoginMessage = ""
	},
}
```

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
+      //edge case: user enters an invalid SRN. Error message is displayed. The user is given another chance. User adds another input field but decided to remove it.
+      // At this point, the variables are reset. So the previously entered SRN is checked again, which should not happen. so setting this validateCount = 2 will bypass this.
+      //Will not affect any other case. The user can either submit or decide to add another field again.
+      this.validateCount = 2;
```
deepansh96, line 212: One solution for this is to keep the state variables for all the input entries each
For example, if the user has entered 3 entries, each one of those entries will have it's own value of `invalidInputMessage` and `invalidLoginMessage` and `validateCount`
That way, while resetting these variables, they'll be reset for one SRN entry only, and other will not be affected

This might take some time so maybe we can park this for later but if you park this for later, make sure you make an issue for this so we don't forget!

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
       list.splice(index, 1);
```
deepansh96: again, why is list needed? Can't you use this.userIDList?

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
         this.invalidLoginMessage = "";
       } else {
         this.invalidInputMessage = "";
       }
+      //if more than 10 characters are entered, slicing the input to only 10
+      // index tells us which input field is being considered
       if (event.target.value.length > this.maxLengthOfSRN) {
         event.target.value = event.target.value.slice(0, this.maxLengthOfSRN);
-        this.userIDList[0]["userID"] = event.target.value;
+        this.userIDList[index]["userID"] = event.target.value;
```
```suggestion
        this.userIDList[index]["userID"] = event.target.value.toString();
```
We have to make sure that when we're storing userIDs in this.userIDList, they are all strings by default

src/components/SRNEntry.vue
```diff
@@ -78,49 +89,73 @@ export default {
...
-    async processForm() {
-      var authType = "SRN";
-      //parsing the userID from user input
-      const userID = parseInt(this.userIDList["0"]["userID"]);
-
+    //method that authentiates the SRN
+    async authenticateSRN(userID) {
       //invokes the validation function
-      let userIsValidated = validateSRN(
+      let userIsValidated = await validateSRN(
```
```suggestion
      let userValidationResponse = await validateSRN(
```

src/components/SRNEntry.vue
```diff
@@ -130,36 +165,52 @@ export default {
...
-              this.redirectTo,
-              this.redirectID,
-              userID,
-              this.isUserValid,
-              authType
-            );
-          }
+      this.isUserValid = userIsValidated.isUserValid;
+      this.validateCount = userIsValidated.validateCount;
+      this.invalidLoginMessage = userIsValidated.invalidLoginMessage;
```
```suggestion
      this.isUserValid = userValidationResponse.isUserValid;
      this.validateCount = userValidationResponse.validateCount;
      this.invalidLoginMessage = userValidationResponse.invalidLoginMessage;
```

src/components/SRNEntry.vue
```diff
@@ -130,36 +165,52 @@ export default {
...
+    },
+    //method called after clicking the submit button
+    async processForm() {
+      var authType = "SRN";
```
deepansh96, line 253: make this a data property no? This value won't change for this component right?

src/components/SRNEntry.vue
```diff
@@ -130,36 +165,52 @@ export default {
...
+      //if only one SRN is entered
+      if (this.userIDList.length == 1) {
+        //parsing the userID from user input
+        let userID = parseInt(this.userIDList["0"]["userID"]);
```
deepansh96: `"0"`?
Does this work? `this.userIDList["0"]`?
I think it needs to be this right? - `this.userIDList[0]`

deepansh96: Also, this line won't be needed if you always store strings in `userID` by default

src/components/SRNEntry.vue
```diff
@@ -130,36 +165,52 @@ export default {
...
+        //invokes the validation function
+        await this.authenticateSRN(userID);
+        //sets the valid flag for that SRN
+        this.userIDList[0]["valid"] = this.isUserValid;
+      } else {
+        //all previously typed SRN's will be authenticated through the addField method.
+        // The last SRN will be authenticated after Submit button is clicked
+        let last = this.userIDList.slice(-1);
+        let userID = parseInt(last[0]["userID"]);
```
deepansh96: again, this will be a one liner if there's a computed variable like `latestEntry`

src/services/API/sqs.js
```diff
@@ -25,8 +25,7 @@ export async function sendSQSMessage(purpose, purposeParams, redirectTo, redirec
         },
         authType: authType,
         user: {
-          values: userID,
-          userDataValidated: isUserValid,
+          values: userIDList,
```
deepansh96, line 28: This means we will send the all the users list at once (in case of multiple SRNs)?

deepansh96 (changes requested):

deepansh96:
Leaving some reviews here because I could not suggest changes in the code

1. The computed property `isAnyUserIDPresent` should be this. Because when you check for `undefined` you automatically check for `null`
```js
isAnyUserIDPresent() {
  //checks if any userID has been typed
  return (
    this.userIDList != undefined &&
    this.userIDList.length > 0 &&
    this.userIDList[0]["userID"] != ""
  );
},
```

2. In the file Homepage.vue, you're computing a variable called `isSingleEntryOnly`. Two suggestions - First is that we should rename it to `areMultipleEntriesAllowed`. This suggestion is mild and you might apply it or not. The second suggestion is that you should compute it in `Homepage.vue` and pass it to the component `SRNEntry.vue`. Currently, `SRNEntry.vue` computes this on it's own. But in future, when we have more types of entries other than SRNEntry, and we also want to figure out if multiple entries are allowed in those components, it would be best if it was passed to them using props from their parent component, which is `Homepage.vue`. So, the code in `Homepage.vue` should be this -
```js
<SRNEntry
  :redirectTo="redirectTo"
  :redirectID="redirectID"
  :purpose="purpose"
  :purposeParams="purposeParams"
  :isSingleEntryOnly="isSingleEntryOnly"
/>
```

3.

src/services/API/sqs.js
sudheshna-donthineni: Yeah, I do the splitting of the SRNs in the lambda.

src/components/SRNEntry.vue
sudheshna-donthineni: Yeah, I'm making that actually a prop from the Homepage component itself. Now that I am building OTP auth, I sending in a prop to each component.

src/components/SRNEntry.vue
sudheshna-donthineni: So, I read that if you want frequent toggling `v-show` is better because the element is already rendered, it just conditionally displays.

src/components/SRNEntry.vue
sudheshna-donthineni: That error didn't come up for me, I wonder why because you are right.

src/components/SRNEntry.vue
sudheshna-donthineni: So what this means is the submit button is disabled until the last SRN that is being typed has 10 characters. length < 9. The index starts from 0. So it is 10 characters only.

src/components/SRNEntry.vue
sudheshna-donthineni: Actually, let me change the name to `isAddButtonAllowed` because that's all it does. Basic validation is in this method. Advanced validation is after the plus button is clicked.

src/components/SRNEntry.vue
```diff
@@ -14,15 +14,17 @@
           pattern="[0-9]*"
           placeholder="Your SRN / आपका SRN"
           required
-          @keypress="isValidSRNFormat($event)"
+          @keypress="isValidNumericEntry($event)"
```
deepansh96, line 34:
```suggestion
          @keypress="allowNumericEntriesOnly($event)"
```

src/components/SRNEntry.vue
```diff
@@ -93,69 +100,98 @@ export default {
...
       //if multiple SRN's are allowed, then this helps with the activation of the + button, to add more SRN's
       return (
         !this.isSingleEntryOnly &&
-        this.userIDList.slice(-1)[0]["userID"].length > this.maxLengthOfSRN - 1 &&
-        this.userIDList.length < 10
+        this.isInputLengthComplete &&
+        this.userIDList.length < numberOfSRNsAllowed
       );
     },
+    isInputLengthComplete() {
```
deepansh96: There is no need for this computed value right?
Whenever you want to use this, you can just use `!this.isCurrentEntryIncomplete` as it will tell us if the current entry is complete or not

src/components/SRNEntry.vue
deepansh96: You haven't tested it? :p

sudheshna-donthineni:
@deepansh96 About the second comment, like I discussed in the call, for next feature, this part comes in as a prop from the Homepage component.
also, added the loading spinner.

deepansh96 (changes requested):

deepansh96 (approved):

vue.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  transpileDependencies: [
+    /@vue\/*/,
+    'vue-router',
+    'plyr',
```
pritamps: plyr? Why?

pritamps (approved):
Approved. but maybe remove `plyr` from the config?

PR #21, opened 2021-10-21
Capturing incorrect SRN entries

### LINKED ISSUE #22: Clear input field when input is incorrect
Problem: When a user types an incorrect SRN, an error message is shown but the input SRN remains in the text field. 
What should be done instead, is clear the text box while showing the error.

### LINKED ISSUE #23: Storing data for incorrect entries as well
Problem: Every time a user types incorrect SRN, the user is displayed with a login error. But we did not have any information about whether the user continued to attempt a second time or even if they did, what the initial attempt was.

sudheshna-donthineni:
- Every time a user types incorrect SRN, the user is displayed with a login error. But we did not have any information about whether the user continued to attempt a second time or even if they did, what the initial attempt was.
- This code does the check every time an SRN is authenticated. Only if it is an invalid entry and if it is the first time, send an SQS message. 
- Added a new `purpose-sub-type` : `'incorrect-entry'`

A minor change: clearing the input field if the input entered is incorrect.

### DIFF
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -129,7 +129,7 @@ export default {
     ifUserEnteredMoreThanOne() {
       return !this.isSingleEntryOnly && this.userIdListLength > 1;
     },
-    //for now, plio does not support multiple input entries
+    /** for now, plio does not support multiple input entries */
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -133,10 +133,13 @@ export default {
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
-    /* returns submit button as disabled if any of the following is true:
-     - no SRN has been typed
-     - input is invalid
-     - SRN hasn't been completely typed */
+    /**
+      * whether the submit button is disabled
+      * true if any of the following conditions are met:
+      * - no SRN has been typed
+      * - input is invalid
+      * - SRN hasn't been completely typed
+      */
     isSubmitButtonDisabled() {
       return (
         !this.isAnyUserIDPresent ||
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -254,14 +254,16 @@ export default {
         this.getLatestEntry["userID"] = "";
       }
 
+      //incorrect entry the first time
       if(!this.isUserValid && this.validateCount == 1){
         var purposeParams = "incorrect-entry"
+        var tempUserIDList = [{userID: userID.toString(), valid: this.isUserValid}]
         sendSQSMessage(
             this.purpose,
             purposeParams,
             this.redirectTo,
             this.redirectID,
-            this.userIDList,
+            tempUserIDList,
             authType
           );
       }
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -54,7 +54,7 @@
     }}</span>
     <span
       class="mx-auto text-red-700 text-base mb-1"
-      v-if="!isUserValid && validateCount == 1"
+      v-if="!isCurrentUserValid && validateCount == 1"
       >{{ invalidLoginMessage }}</span
     >
     <!-- add srn button -->
@@ -105,7 +105,7 @@ export default {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
       invalidInputMessage: null, //flag to check if the input is in correct format
-      isUserValid: false, // whether the user exists in the backend database
+      isCurrentUserValid: false, // temp flag 
       maxLengthOfSRN: 10,
       validateCount: 0, //this variable tells us how many times the user has been validated.
       invalidLoginMessage: "Please enter correct SRN / कृपया सही SRN दर्ज करें",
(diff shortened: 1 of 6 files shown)


deepansh96:
@sudheshna-donthineni Just one feedback,

The way you outlined the need for the PR and the things that are changed in this PR at the very top is perfect. Every PR should have details like this so anyone who's reviewing has instant context but I think it would be even more helpful to put all those details in an issue and then link the issue here. 

For example, two issues would be created here. One for adding a new purpose type of `incorrect-entry` and other issue for clearing the input filed if input entered is incorrect.

How this will help? At any time in the future if we want to see what things were fixed, we won't have to iterate through every pr and open and see the description. We would have listed everything in the issues tab

src/components/SRNEntry.vue
```diff
@@ -247,10 +248,29 @@ export default {
       this.validateCount = userValidationResponse.validateCount;
       this.invalidLoginMessage = userValidationResponse.invalidLoginMessage;
       this.isLoading = false;
+
+      //clear the input field if entry is incorrect
+      if(this.invalidLoginMessage != ""){
+        this.getLatestEntry["userID"] = "";
```
deepansh96: Suggestion --
Make a method, something like, `resetEntry(entryIndex)` which would take in an argument index, and reset the value of the input box at that index.

This would help us to have a standard method to reset any input box.

src/components/SRNEntry.vue
```diff
@@ -247,10 +248,29 @@ export default {
...
+      }
+
+      //incorrect entry the first time
+      if(!this.isUserValid && this.validateCount == 1){
+        var purposeParams = "incorrect-entry"
+        var tempUserIDList = [{userID: userID.toString(), valid: this.isUserValid}]
```
deepansh96: Why do you need a temp list here?

src/components/SRNEntry.vue
```diff
@@ -91,6 +91,7 @@ import { redirectToDestination } from "@/services/redirectToDestination.js";
 import { sendSQSMessage } from "@/services/API/sqs";
 
 const numberOfSRNsAllowed = 10;
+const authType = "SRN";
```
deepansh96, line 92: I'm sorry I think I asked you to make this a global const, but it's better to make it a data property. The data property object is made to store data that needs to be used across the component.

src/components/SRNEntry.vue
```diff
@@ -247,10 +248,29 @@ export default {
...
+        var tempUserIDList = [{userID: userID.toString(), valid: this.isUserValid}]
+        sendSQSMessage(
+            this.purpose,
+            purposeParams,
+            this.redirectTo,
+            this.redirectID,
+            tempUserIDList,
+            authType
+          );
+      }
```
deepansh96: I think this is not the right place for this logic.
You need to run this piece of logic when someone either submits (after entering a wrong srn) or clicks "add another srn" (after entering a wrong srn).

2 reasons why this logic should not be here.
- This method is called `authenticateSRN` and it's job should just to be do the authentication and setting the respective variables. After setting the needed variables, it should not care about what to do next.
- You need the above logic to run in two cases - clicking submit and clicking add srn button. So this new logic should be present in those two functions `processForm` and `addField`. And to write the same logic two times in two functions would be redundant, you should make a separate function, something on the lines of, `handleIncorrectEntry()` and use that where you need to.

I hope it was clear enough? Let me know if you have any doubts

deepansh96 (changes requested):

src/components/SRNEntry.vue
sudheshna-donthineni: Actually, I let this be here because it'll soon be turned into a prop. I thought of doing it for this PR, but that would also change the URL format. So will do it along with the OTP auth component.

src/components/SRNEntry.vue
sudheshna-donthineni: because the message being sent is specific to each userID. And this is the format the lambda handler parses the input. If I just send the userID, then I'll need to add a whole other logic for the lambda.

src/components/SRNEntry.vue
sudheshna-donthineni: Actually, `authenticateSRN` is being run in both the logic, so this piece of code also will run both times. But yeah, it makes sense that this method shouldn't worry about what is being done to the result.

src/components/SRNEntry.vue
deepansh96: Cool cool

src/components/SRNEntry.vue
```diff
@@ -187,11 +187,16 @@ export default {
...
+    //resets input field at an index
+    resetEntry(index){
+        this.userIDList[index]["userID"] = "";
+    },
     async addField() {
       //for adding another field, the previously entered ID is validated against the database
       const latestUserID = parseInt(this.getLatestEntry["userID"]);
       if (!isNaN(latestUserID)) {
         await this.authenticateSRN(latestUserID);
+        this.handleIncorrectEntry(latestUserID);
```
deepansh96: Looks good! Just one thing - if someone is reading the code of `addField` or the code of `processForm`, they will notice that after the authentication is done, incorrect entries are handled be default.

I think it'll make much more sense if we call `handleIncorrectEntry` after some if condition. 
Basically, in plain, English
```
authenticate
if (the current entry is invalid)
then this.handleIncorrectEntry
```

deepansh96 (changes requested):

pritamps (commented):
This looks good to me! With the change @deepansh96 has suggested, will be good to merge!

src/components/SRNEntry.vue
```diff
@@ -235,9 +237,8 @@ export default {
         this.userIDList[index]["userID"] = event.target.value.toString();
       }
     },
+    //incorrect entry the first time
```
deepansh96: I don't think this comment is valid here right? This method handles incorrect entries and that's it. This doesn't know or care if the entry is being made the first time or nth time.
We care about that in the parent methods where this method is being called

Instead of this comment, I would suggest writing how ANY incorrect entry is handled
Also I think we should get into the habit of writing multi line comments!
For example, the below comment tells us what the function does, and if there are any parameters that need to be passed to the function.
```js
/**
 * Comment goes here - what does the function do?
 * @param {Number} - param1Name
 * @param {String} - param2Name 
 */
```
These comments should be put on top of a function. No need of multiline comments inside a function

Also, no need to change all of them now, but slowly slowly we should convert all our function docstrings into multi line docstrings

deepansh96 (changes requested):

PR #24, opened 2021-11-08
added detailed comments

sudheshna-donthineni:
Added detailed comments to the entire code.

### DIFF
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -129,7 +129,7 @@ export default {
     ifUserEnteredMoreThanOne() {
       return !this.isSingleEntryOnly && this.userIdListLength > 1;
     },
-    //for now, plio does not support multiple input entries
+    /** for now, plio does not support multiple input entries */
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -133,10 +133,13 @@ export default {
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
-    /* returns submit button as disabled if any of the following is true:
-     - no SRN has been typed
-     - input is invalid
-     - SRN hasn't been completely typed */
+    /**
+      * whether the submit button is disabled
+      * true if any of the following conditions are met:
+      * - no SRN has been typed
+      * - input is invalid
+      * - SRN hasn't been completely typed
+      */
     isSubmitButtonDisabled() {
       return (
         !this.isAnyUserIDPresent ||
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -54,7 +54,7 @@
     }}</span>
     <span
       class="mx-auto text-red-700 text-base mb-1"
-      v-if="!isUserValid && validateCount == 1"
+      v-if="!isCurrentUserValid && validateCount == 1"
       >{{ invalidLoginMessage }}</span
     >
     <!-- add srn button -->
@@ -105,7 +105,7 @@ export default {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
       invalidInputMessage: null, //flag to check if the input is in correct format
-      isUserValid: false, // whether the user exists in the backend database
+      isCurrentUserValid: false, // temp flag 
       maxLengthOfSRN: 10,
       validateCount: 0, //this variable tells us how many times the user has been validated.
       invalidLoginMessage: "Please enter correct SRN / कृपया सही SRN दर्ज करें",
@@ -113,48 +113,63 @@ export default {
     };
   },
   computed: {
+    //returns length of the userID list
     userIdListLength() {
       return this.userIDList.length;
     },
+    //checks if any userID has been typed
     isAnyUserIDPresent() {
-      //checks if any userID has been typed
       return (
         this.userIDList != undefined &&
         this.userIdListLength > 0 &&
         this.userIDList[0]["userID"] != ""
       );
     },
+    //checks if multiple entries are allowed and if one entry is already inputted
     ifUserEnteredMoreThanOne() {
-      //used when multiple SRN's will be allowed to typed
       return !this.isSingleEntryOnly && this.userIdListLength > 1;
     },
+    //for now, plio does not support multiple input entries
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
+    /* returns submit button as disabled if any of the following is true:
+     - no SRN has been typed
+     - input is invalid
+     - SRN hasn't been completely typed */
     isSubmitButtonDisabled() {
-      //submit button is disabled if no SRN has been typed or input is invalid or SRN hasn't been completely typed
       return (
         !this.isAnyUserIDPresent ||
         this.invalidInputMessage != "" ||
         this.isCurrentEntryIncomplete
       );
     },
+    /* checks if + button should be displayed. Will be activated only if:
+     - multiple entries are allowed
+     - if current input entry is complete
+     - if cap of maximum entries hasn't been reached yet */
     isAddButtonAllowed() {
-      //if multiple SRN's are allowed, then this helps with the activation of the + button, to add more SRN's
       return (
         !this.isSingleEntryOnly &&
         !this.isCurrentEntryIncomplete &&
         this.userIdListLength < numberOfSRNsAllowed
       );
     },
+    // checks if the current input entry has the required number of characters.
     isCurrentEntryIncomplete() {
       return this.getLatestEntry["userID"].length < this.maxLengthOfSRN;
     },
+    // returns the most recent typed input entry
     getLatestEntry() {
       return this.userIDList.slice(-1)[0];
     },
   },
   methods: {
+    /* determines how the input box should look like depending on which box it is. 
+      - if the input box is the current one (i.e. being typed) and if the input is invalid (invalid input message is being displayed),
+        then, the box has a border of red
+      - if the input box is a past one (i.e. already typed), then the box is shown as unclickable 
+      @param {Number} - index - index of the input box */
     calculateInputboxStyleClasses(index) {
       return [
         {
@@ -164,64 +179,87 @@ export default {
         },
       ];
     },
+    /* checks to see if the input character is a number. Makes use of ASCII values.
+    @param {Event} - e - event triggered when a character is typed.  */
     isValidNumericEntry(e) {
-      //checking to see if each char typed by user is only a number
       if (e.keyCode >= 48 && e.keyCode <= 57) return true;
       else e.preventDefault();
     },
+    // adds a new dict to the array - userIDList
     addNewEmptyField() {
       this.userIDList.push({ userID: "", valid: false });
     },
+    /* removes an element in the array at a given index. 
+    @param {Number} - index - the index of the input box where - button is clicked. */
     removeInputField(index) {
       this.userIDList.splice(index, 1);
     },
+    // sets the invalid input message to null (default)
     resetInvalidInputMessage() {
       this.invalidInputMessage = "";
     },
+    // sets the invalid login message to null (default)
     resetInvalidLoginMessage() {
       this.invalidLoginMessage = "";
     },
+    // sets the temp valid flag to false (default) for authentication of a possible next user
     resetValidFlag() {
-      this.isUserValid = false;
+      this.isCurrentUserValid = false;
     },
+    // sets the valid key of the latest user to the value of the temp flag. The temp flag contains the value returned by the backend
     setValidFlag() {
-      this.getLatestEntry["valid"] = this.isUserValid;
+      this.getLatestEntry["valid"] = this.isCurrentUserValid;
     },
-    //resets input field at an index
+    //resets the userID key of an element in the array at a particular index.
     resetEntry(index){
         this.userIDList[index]["userID"] = "";
     },
+    /* this function is called whenever the + button is clicked. The most recent typed entry is authenticated against the database. 
+    - if it does not exist and it is the first time being authenticated, handleIncorrectEntry() is called. 
+    - if the entry exists or if it is being authenticated the second time, the 'valid' flag of the entry is set to whatever the value is returned by the backend. 
+      Also, isCurrentUserValid and validateCount are reset to their default values, so that a possible next entry can be processed. */
     async addField() {
-      //for adding another field, the previously entered ID is validated against the database
       const latestUserID = parseInt(this.getLatestEntry["userID"]);
       if (!isNaN(latestUserID)) {
         await this.authenticateSRN(latestUserID);
-        if(!this.isUserValid && this.validateCount == 1){
+        if(!this.isCurrentUserValid && this.validateCount == 1){
           this.handleIncorrectEntry(latestUserID);
         }
-        //only if the SRN is valid or if the user is entering a SRN for the second time,the loop is entered
-        if (this.isUserValid || this.validateCount > 1) {
-          //setting the flag of the SRN
+        if (this.isCurrentUserValid || this.validateCount > 1) {
           this.setValidFlag();
           this.addNewEmptyField();
-          //resetting flags to default for processing next SRN
           this.resetValidFlag;
           this.validateCount = 0;
         }
       }
     },
+    /* Before an element is removed from the userIDList array, the login message and input message are reset to default values. 
+       removeInputField() is called to remove the element. 
+       In most cases, this would be enough. But there is one edge case:
+      Let's say the user typed in ID 'A' and decides to add another entry. This means 'A' is authenticated and it is valid/invalid. 
+      The user types in the ID 'B' as the next entry and it is invalid. 
+      The user changed their mind and decided to delete this entry ('B'). At this point, the variables will be in the following state:
+        validateCount = 1
+        isCurrentUserValid = false
+      The state of the variables contradict the fact that 'A' is already authenticated. This means any of the two cases:
+      - A is valid, so validateCount = 0 and isCurrentUserValid = true
+      - A is invalid, so validateCount = 2 and isCurrentUserValid = false
+      So when the user clicks submit, 'A' is authenticated again, which is a behaviour we do not want. So to bypass this, validateCount is set to 2. 
+      (To better understand how this will help, understanding processForm() will help.) 
+      isCurrentUserValid flag is ignored because the value is already set in each user's valid flag. 
+      @params {Number} - index - index of input field to be removed */
     removeField(index) {
-      //resetting all messages to default before deleting the input field
       this.resetInvalidInputMessage;
       this.resetInvalidLoginMessage;
-      //edge case: user enters an invalid SRN. Error message is displayed. The user is given another chance. User adds another input field but decided to remove it.
-      // At this point, the variables are reset. So the previously entered SRN is checked again, which should not happen. so setting this validateCount = 2 will bypass this.
-      //Will not affect any other case. The user can either submit or decide to add another field again.
       this.validateCount = 2;
       this.removeInputField(index);
     },
+    /*  this function is called whenever something is entered in the input box. It checks if the required number of characters are being typed. 
+    Until then, it prompts the user to type the required characters. 
+    If the user types more than the desired number of characters, the input is sliced and the user wont be able to see the extra characters.
+    @params {Event} - event - the event which triggered this function
+    @params {Number} - index - the index of the input field */
     updateValue(event, index) {
-      //checks if the 10 characters are entered
       if (event.target.value.length == 0) {
         this.invalidInputMessage = "";
       } else if (event.target.value.length < this.maxLengthOfSRN) {
@@ -230,17 +268,22 @@ export default {
       } else {
         this.resetInvalidInputMessage();
       }
-      //if more than 10 characters are entered, slicing the input to only 10
-      // index tells us which input field is being considered
       if (event.target.value.length > this.maxLengthOfSRN) {
         event.target.value = event.target.value.slice(0, this.maxLengthOfSRN);
         this.userIDList[index]["userID"] = event.target.value.toString();
       }
     },
-    //incorrect entry the first time
+    /* this functions handle all invalid/incorrect entries. An SQS message is sent with the following parameters:
+    - purpose - from URL
+    - purposeParams - "incorrect-entry"
+    - redirectTo - from URL
+    - redirectID - from URL
+    - tempUserIDList - which contains only the ID of the incorrect entry
+    - authType - from URL 
+    @params {String} - userID - ID of the incorrect entry field */
     handleIncorrectEntry(userID){
         var purposeParams = "incorrect-entry"
-        var tempUserIDList = [{userID: userID.toString(), valid: this.isUserValid}]
+        var tempUserIDList = [{userID: userID.toString(), valid: this.isCurrentUserValid}]
         sendSQSMessage(
             this.purpose,
             purposeParams,
@@ -250,50 +293,45 @@ export default {
             authType
           );
     },
-    //method that authentiates the SRN
+    /* this method is called whenever + button is clicked. It authenticates the most recent typed ID. 
+    If the ID is invalid, then the login mesasge displays an error and the input field is cleared for the user to correct their entry.
+    @params {String} - userID - most recent ID  */
     async authenticateSRN(userID) {
       this.isLoading = true;
-      //invokes the validation function
       let userValidationResponse = await validateSRN(
         userID,
         this.validateCount,
         this.isSingleEntryOnly,
         this.redirectID,
-        this.isUserValid,
+        this.isCurrentUserValid,
         this.purpose,
         this.purposeParams,
         this.redirectTo
       );
-      this.isUserValid = userValidationResponse.isUserValid;
+      this.isCurrentUserValid = userValidationResponse.isCurrentUserValid;
       this.validateCount = userValidationResponse.validateCount;
       this.invalidLoginMessage = userValidationResponse.invalidLoginMessage;
       this.isLoading = false;
-      //clear the input field if entry is incorrect
+
       if(this.invalidLoginMessage != ""){
         this.resetEntry(this.userIdListLength - 1)
       }
-      
-
-      
     },
-    //method called after clicking the submit button
+    /* this method is called after the user clicks the submit button. Since all the ID's but the last are authenticated, this method calls authenticateSRN(). 
+    If the user is invalid and is being authenticated the first time, handleIncorrectEntry() is called. 
+    If the user is valid or is being authenticated the second time, the user is reidrected to their destination and an SQS message is sent. */
     async processForm() {
-   
-
-      //all previously typed SRN's will be authenticated through the addField method.
-      // The last SRN will be authenticated after Submit button is clicked
-      // (will work even for single entry as the first entry can be also considered as the last entry)
       let latestUserID = parseInt(this.getLatestEntry["userID"]);
       if (!isNaN(latestUserID)) {
         await this.authenticateSRN(latestUserID);
-        if(!this.isUserValid && this.validateCount == 1){
+        if(!this.isCurrentUserValid && this.validateCount == 1){
         this.handleIncorrectEntry(latestUserID);
         }
         this.setValidFlag();
       }
 
       // either the user is valid or the user has been checked twice
-      if (this.isUserValid || this.validateCount > 1) {
+      if (this.isCurrentUserValid || this.validateCount > 1) {
         if (
           redirectToDestination(
             this.purposeParams,
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -113,7 +113,7 @@ export default {
     };
   },
   computed: {
-    //returns length of the userID list
+    /** returns length of the list of user IDs */
     userIdListLength() {
       return this.userIDList.length;
     },
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -105,7 +105,7 @@ export default {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
       invalidInputMessage: null, //flag to check if the input is in correct format
-      isCurrentUserValid: false, // temp flag which contains the validity of the most recent user ID
+      isCurrentUserValid: false, // whether the current user is valid
       maxLengthOfSRN: 10,
       validateCount: 0, //this variable tells us how many times the user has been validated.
       invalidLoginMessage: "Please enter correct SRN / कृपया सही SRN दर्ज करें",
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -162,8 +162,8 @@ export default {
     isCurrentEntryIncomplete() {
       return this.getLatestEntry["userID"].length < this.maxLengthOfSRN;
     },
-    // returns the most recent typed input entry
-    getLatestEntry() {
+    // returns the most recently entered input
+    latestEntry() {
       return this.userIDList.slice(-1)[0];
     },
   },
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -260,7 +260,7 @@ export default {
     /*  this function is called whenever something is entered in the input box. It checks if the required number of characters are being typed. 
     Until then, it prompts the user to type the required characters. 
     If the user types more than the desired number of characters, the input is sliced and the user wont be able to see the extra characters.
-    @params {Event} - event - the event which triggered this function
+    @param {Object} - event - the event which triggered this function
     @params {Number} - index - the index of the input field */
     updateValue(event, index) {
       if (event.target.value.length == 0) {
--- a/src/components/SRNEntry.vue
+++ b/src/components/SRNEntry.vue
@@ -104,7 +104,7 @@ export default {
   data() {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
-      invalidInputMessage: null, //flag to check if the input is in correct format
+      invalidInputMessage: null, // whether the input is in correct format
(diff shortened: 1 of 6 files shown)


src/components/SRNEntry.vue
```diff
@@ -54,7 +54,7 @@
     }}</span>
     <span
       class="mx-auto text-red-700 text-base mb-1"
-      v-if="!isUserValid && validateCount == 1"
+      v-if="!isCurrentUserValid && validateCount == 1"
```
dalmia: might be better (in terms of readability) to wrap this within a computed property with a name (like): `isInvalidLoginMessageShown`

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
       invalidInputMessage: null, //flag to check if the input is in correct format
-      isUserValid: false, // whether the user exists in the backend database
+      isCurrentUserValid: false, // temp flag which contains the validity of the most recent user ID
```
dalmia: maybe simpler?
```suggestion
      isCurrentUserValid: false, // whether the current user is valid
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
       maxLengthOfSRN: 10,
       validateCount: 0, //this variable tells us how many times the user has been validated.
       invalidLoginMessage: "Please enter correct SRN / कृपया सही SRN दर्ज करें",
-      isLoading: false,
+      isLoading: false, 
     };
   },
   computed: {
+    //returns length of the userID list
```
dalmia:
```suggestion
    /** returns length of the list of user IDs */
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
     userIdListLength() {
       return this.userIDList.length;
     },
+    //checks if any userID has been typed
```
dalmia: `/**  */` is used for docstrings
```suggestion
    /** checks if any userID has been entered */
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
         this.userIdListLength > 0 &&
         this.userIDList[0]["userID"] != ""
       );
     },
+    //checks if multiple entries are allowed and if one entry is already inputted
     ifUserEnteredMoreThanOne() {
-      //used when multiple SRN's will be allowed to typed
       return !this.isSingleEntryOnly && this.userIdListLength > 1;
     },
+    //for now, plio does not support multiple input entries
```
dalmia: would be better to first mention **what** does this check and then mention **why** this check is being done  
```suggestion
    /** for now, plio does not support multiple input entries */
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
+    //checks if any userID has been typed
     isAnyUserIDPresent() {
-      //checks if any userID has been typed
       return (
         this.userIDList != undefined &&
         this.userIdListLength > 0 &&
         this.userIDList[0]["userID"] != ""
       );
     },
+    //checks if multiple entries are allowed and if one entry is already inputted
```
dalmia: the question of whether it is allowed or not is one condition for evaluating this result - doesn't need to be mentioned in the docstring
```suggestion
    /** whether multiple entries have been made by the user */
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
     },
+    //for now, plio does not support multiple input entries
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
+    /* returns submit button as disabled if any of the following is true:
+     - no SRN has been typed
+     - input is invalid
+     - SRN hasn't been completely typed */
     isSubmitButtonDisabled() {
```
dalmia:
```suggestion
    /**
      * whether the submit button is disabled
      * true if any of the following conditions are met:
      * - no SRN has been typed
      * - input is invalid
      * - SRN hasn't been completely typed
      */
    isSubmitButtonDisabled() {
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
       return (
         !this.isAnyUserIDPresent ||
         this.invalidInputMessage != "" ||
         this.isCurrentEntryIncomplete
       );
     },
+    /* checks if + button should be displayed. Will be activated only if:
+     - multiple entries are allowed
+     - if current input entry is complete
+     - if cap of maximum entries hasn't been reached yet */
```
dalmia: change the formatting of the docstring similar to how it is being done for `isSubmitButtonDisabled`

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
     isAddButtonAllowed() {
-      //if multiple SRN's are allowed, then this helps with the activation of the + button, to add more SRN's
       return (
         !this.isSingleEntryOnly &&
         !this.isCurrentEntryIncomplete &&
         this.userIdListLength < numberOfSRNsAllowed
       );
     },
+    // checks if the current input entry has the required number of characters.
```
dalmia: use `/** */` everywhere for the docstrings

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
     isCurrentEntryIncomplete() {
       return this.getLatestEntry["userID"].length < this.maxLengthOfSRN;
     },
+    // returns the most recent typed input entry
```
dalmia: use `/** */` everywhere for the docstrings

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
     getLatestEntry() {
```
dalmia: `get...` is a name for a method. Since this is a computed property, the more appropriate name is `latestEntry`:

```suggestion
    // returns the most recently entered input
    latestEntry() {
```

src/services/API/checkUser.js
```diff
@@ -2,10 +2,10 @@ import {client} from "@/services/API/rootClient.js"
 import { checkUserEndpoint } from "@/services/API/endpoints.js";
 
 export default {
-    //checks if the user exists in the database or not
+    /* this method validates the ID against the database
+    @params {String} inputUserID - each ID being typed*/
```
dalmia:
```suggestion
    /**
      * validates that the ID exists
      * @param {String} inputUserID - the id that needs to be validated
      */
```

src/services/API/sqs.js
```diff
@@ -10,6 +10,15 @@ const sqsClient = new SQSClient({
...
 
+/* This function is used to send an SQS message to the queue on AWS. Makes use of the AWS SDK. (SendMessageCommand)
+@params {String} - purpose
+@params {String} - purposeParams
+@params {String} - redirectTo
+@params {String} - redirectID
+@params {Array} - userIDList - list of users wanting to go through the layer
+@params {String} - authType
+Everything, except the userIDList, is extracted from the auth layer URL
+*/
```
dalmia: `@param` not `@params`
the format is:
`@param {type} {variable name} - {description}`
please remove the `-` between `{type}` and `{variable name}`.

+ format the docstring according to the feedback given before
+ add descriptions for each param
+ no need to mention implementation details like `SendMessageCommand`  in the docstring

src/services/redirectToDestination.js
```diff
@@ -1,33 +1,42 @@
...
+        - queryparams: Plio's API key (stored in environment variable) and unique ID (user ID) 
+- if the destination is a google meet class, the url is built with the following:
+        - the meet's base URL (stored in environment variable)
+        - the meeting ID (from auth layer URL, redirectID) 
+- if the destination is none of the above, then an SQS message with purpose = 'Error' is sent to keep a log of.
+@params {String} purposeparams - extracted from auth layer URL
+@params {Array} userIDList - list of userID's wanting to go through the layer 
+@params {String} redirectID - extracted from auth layer URL
+@params {String} redirectTo - extracted from auth layer URL
+@params {String} authType - extracted from auth layer URL */
```
dalmia: incorporate the feedback on docstrings given before.

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
     setValidFlag() {
-      this.getLatestEntry["valid"] = this.isUserValid;
+      this.getLatestEntry["valid"] = this.isCurrentUserValid;
     },
-    //resets input field at an index
+    //resets the userID key of an element in the array at a particular index.
     resetEntry(index){
         this.userIDList[index]["userID"] = "";
     },
+    /* this function is called whenever the + button is clicked. The most recent typed entry is authenticated against the database.
```
pritamps: Function comments should be after the function name (next line)

pritamps (changes requested):
Start all function docstrings with capital letter, move function docstrings to line after function name.

src/services/validation.js
```diff
@@ -1,34 +1,35 @@
...
+            isCurrentUserValid: isCurrentUserValid,
             validateCount: validateCount,
             invalidLoginMessage: invalidLoginMessage
             }
         }
 
-
-    // this condition checks if the user is getting authenticated the first time. Just shows an error message.
-    if (!isUserValid && validateCount == 0) {
+    if (!isCurrentUserValid && validateCount == 0) {
```
dalmia: you no longer need to check for `!isCurrentUserValid` right? otherwise the code would have stopped in the earlier return statement itself.

src/services/validation.js
```diff
@@ -1,34 +1,35 @@
...
-    
+/* this function validates the entry against the database (for this usecase, it is firestore). 
+firebaseAPI service is used to validate and it returns a boolean value, indicating whether the user is valid or not.
+If the user is valid and validateCount = 0, then the function returns.
+If the user is not valid and validateCount = 0, then the user is not valid, so validateCount becomes 1 and login message displays an error, allowing the user to correct their entry.
+If the user is not valid and validateCount = 1, then the user has already been authenticated once before and the user has retyped, yet the entry is still invalid, then validateCount becomes 2 to indicate this behaviour.
+@params {String} - userID - current ID being authenticated 
+@params {Number} - validateCount - indicates how many times the user has been validated 
+@params {Boolean} - isCurrentUserValid - holds the value returned by the API */
+export async function validateSRN(userID, validateCount, isCurrentUserValid){
```
dalmia: why is `isCurrentUserValid` being passed into the function when the value is being overwritten by what the API returns?

src/services/validation.js
```diff
@@ -1,34 +1,35 @@
...
-    // validateCount = 0, when the user gets validated on the first try
-    // validateCount = 1, when the user fails the first check
-    // validateCount = 2, when user fails the second check as well -> letting the user go through irrespective.
-    
+/* this function validates the entry against the database (for this usecase, it is firestore). 
+firebaseAPI service is used to validate and it returns a boolean value, indicating whether the user is valid or not.
+If the user is valid and validateCount = 0, then the function returns.
+If the user is not valid and validateCount = 0, then the user is not valid, so validateCount becomes 1 and login message displays an error, allowing the user to correct their entry.
+If the user is not valid and validateCount = 1, then the user has already been authenticated once before and the user has retyped, yet the entry is still invalid, then validateCount becomes 2 to indicate this behaviour.
+@params {String} - userID - current ID being authenticated
```
dalmia:
```suggestion
@params {String} - userID - the ID that needs to be validated 
```

src/services/validation.js
dalmia: use the docstring format and suggestions given in the earlier comments.

src/components/SRNEntry.vue
```diff
@@ -250,50 +293,45 @@ export default {
...
         this.resetEntry(this.userIdListLength - 1)
       }
-      
-
-      
     },
-    //method called after clicking the submit button
+    /* this method is called after the user clicks the submit button. Since all the ID's but the last are authenticated, this method calls authenticateSRN(). 
+    If the user is invalid and is being authenticated the first time, handleIncorrectEntry() is called. 
+    If the user is valid or is being authenticated the second time, the user is reidrected to their destination and an SQS message is sent. */
```
dalmia: I don't see why you need to add the implementation details in the docstring - about which methods to call in which situations. The docstring is meant to give a high-level overview of what the function does!

+ please follow the docstring format suggested above.

src/components/SRNEntry.vue
```diff
@@ -250,50 +293,45 @@ export default {
             authType
           );
     },
-    //method that authentiates the SRN
+    /* this method is called whenever + button is clicked. It authenticates the most recent typed ID. 
+    If the ID is invalid, then the login mesasge displays an error and the input field is cleared for the user to correct their entry.
+    @params {String} - userID - most recent ID  */
```
dalmia: same comments on docstrings as before + typos

src/components/SRNEntry.vue
```diff
@@ -230,17 +268,22 @@ export default {
...
     },
-    //incorrect entry the first time
+    /* this functions handle all invalid/incorrect entries. An SQS message is sent with the following parameters:
+    - purpose - from URL
+    - purposeParams - "incorrect-entry"
+    - redirectTo - from URL
+    - redirectID - from URL
+    - tempUserIDList - which contains only the ID of the incorrect entry
+    - authType - from URL 
+    @params {String} - userID - ID of the incorrect entry field */
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
-      // At this point, the variables are reset. So the previously entered SRN is checked again, which should not happen. so setting this validateCount = 2 will bypass this.
-      //Will not affect any other case. The user can either submit or decide to add another field again.
       this.validateCount = 2;
       this.removeInputField(index);
     },
+    /*  this function is called whenever something is entered in the input box. It checks if the required number of characters are being typed. 
+    Until then, it prompts the user to type the required characters. 
+    If the user types more than the desired number of characters, the input is sliced and the user wont be able to see the extra characters.
+    @params {Event} - event - the event which triggered this function
+    @params {Number} - index - the index of the input field */
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
+      The user changed their mind and decided to delete this entry ('B'). At this point, the variables will be in the following state:
+        validateCount = 1
+        isCurrentUserValid = false
+      The state of the variables contradict the fact that 'A' is already authenticated. This means any of the two cases:
+      - A is valid, so validateCount = 0 and isCurrentUserValid = true
+      - A is invalid, so validateCount = 2 and isCurrentUserValid = false
+      So when the user clicks submit, 'A' is authenticated again, which is a behaviour we do not want. So to bypass this, validateCount is set to 2. 
+      (To better understand how this will help, understanding processForm() will help.) 
+      isCurrentUserValid flag is ignored because the value is already set in each user's valid flag. 
+      @params {Number} - index - index of input field to be removed */
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
     },
-    //resets input field at an index
+    //resets the userID key of an element in the array at a particular index.
     resetEntry(index){
         this.userIDList[index]["userID"] = "";
     },
+    /* this function is called whenever the + button is clicked. The most recent typed entry is authenticated against the database. 
+    - if it does not exist and it is the first time being authenticated, handleIncorrectEntry() is called. 
+    - if the entry exists or if it is being authenticated the second time, the 'valid' flag of the entry is set to whatever the value is returned by the backend. 
+      Also, isCurrentUserValid and validateCount are reset to their default values, so that a possible next entry can be processed. */
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
     getLatestEntry() {
       return this.userIDList.slice(-1)[0];
     },
   },
   methods: {
+    /* determines how the input box should look like depending on which box it is. 
+      - if the input box is the current one (i.e. being typed) and if the input is invalid (invalid input message is being displayed),
+        then, the box has a border of red
+      - if the input box is a past one (i.e. already typed), then the box is shown as unclickable 
+      @param {Number} - index - index of the input box */
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
         },
       ];
     },
+    /* checks to see if the input character is a number. Makes use of ASCII values.
+    @param {Event} - e - event triggered when a character is typed.  */
     isValidNumericEntry(e) {
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
-      //checking to see if each char typed by user is only a number
       if (e.keyCode >= 48 && e.keyCode <= 57) return true;
       else e.preventDefault();
     },
+    // adds a new dict to the array - userIDList
     addNewEmptyField() {
       this.userIDList.push({ userID: "", valid: false });
     },
+    /* removes an element in the array at a given index. 
+    @param {Number} - index - the index of the input box where - button is clicked. */
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
       ];
     },
+    /* checks to see if the input character is a number. Makes use of ASCII values.
+    @param {Event} - e - event triggered when a character is typed.  */
     isValidNumericEntry(e) {
-      //checking to see if each char typed by user is only a number
       if (e.keyCode >= 48 && e.keyCode <= 57) return true;
       else e.preventDefault();
     },
+    // adds a new dict to the array - userIDList
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
     addNewEmptyField() {
       this.userIDList.push({ userID: "", valid: false });
     },
+    /* removes an element in the array at a given index. 
+    @param {Number} - index - the index of the input box where - button is clicked. */
     removeInputField(index) {
       this.userIDList.splice(index, 1);
     },
+    // sets the invalid input message to null (default)
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
     resetInvalidInputMessage() {
       this.invalidInputMessage = "";
     },
+    // sets the invalid login message to null (default)
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
-      this.isUserValid = false;
+      this.isCurrentUserValid = false;
     },
+    // sets the valid key of the latest user to the value of the temp flag. The temp flag contains the value returned by the backend
     setValidFlag() {
-      this.getLatestEntry["valid"] = this.isUserValid;
+      this.getLatestEntry["valid"] = this.isCurrentUserValid;
     },
-    //resets input field at an index
+    //resets the userID key of an element in the array at a particular index.
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
     },
+    // sets the invalid input message to null (default)
     resetInvalidInputMessage() {
       this.invalidInputMessage = "";
     },
+    // sets the invalid login message to null (default)
     resetInvalidLoginMessage() {
       this.invalidLoginMessage = "";
     },
+    // sets the temp valid flag to false (default) for authentication of a possible next user
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
     resetValidFlag() {
-      this.isUserValid = false;
+      this.isCurrentUserValid = false;
     },
+    // sets the valid key of the latest user to the value of the temp flag. The temp flag contains the value returned by the backend
```
dalmia: same comments on docstrings as before

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-id's and a valid flag for each
       invalidInputMessage: null, //flag to check if the input is in correct format
```
```suggestion
      invalidInputMessage: null, // whether the input is in correct format
```

src/components/SRNEntry.vue
```diff
@@ -105,56 +105,71 @@ export default {
...
-      isUserValid: false, // whether the user exists in the backend database
+      isCurrentUserValid: false, // temp flag which contains the validity of the most recent user ID
       maxLengthOfSRN: 10,
       validateCount: 0, //this variable tells us how many times the user has been validated.
       invalidLoginMessage: "Please enter correct SRN / कृपया सही SRN दर्ज करें",
-      isLoading: false,
+      isLoading: false,
```
deepansh96: I think your linter is not working maybe?
Because in my and Aman's system, whenever we save a `.vue` file, it automatically lints everything.
So this extra space that has been added, should've been removed automatically

Can you check once if your linter is working fine?

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
         },
       ];
     },
+    /* checks to see if the input character is a number. Makes use of ASCII values.
+    @param {Event} - e - event triggered when a character is typed.  */
```
```suggestion
    @param {Object} - e - event triggered when a character is typed.  */
```
There's no data type called as `Event` right?
It's basically an Object that's being passed into `e`

src/components/SRNEntry.vue
deepansh96: Do you mean docstrings?
Docstrings in JS are typed just above the function name

src/components/SRNEntry.vue
```diff
@@ -164,64 +179,87 @@ export default {
...
-      //edge case: user enters an invalid SRN. Error message is displayed. The user is given another chance. User adds another input field but decided to remove it.
-      // At this point, the variables are reset. So the previously entered SRN is checked again, which should not happen. so setting this validateCount = 2 will bypass this.
-      //Will not affect any other case. The user can either submit or decide to add another field again.
       this.validateCount = 2;
       this.removeInputField(index);
     },
+    /*  this function is called whenever something is entered in the input box. It checks if the required number of characters are being typed. 
+    Until then, it prompts the user to type the required characters. 
+    If the user types more than the desired number of characters, the input is sliced and the user wont be able to see the extra characters.
+    @params {Event} - event - the event which triggered this function
```
```suggestion
    @param {Object} - event - the event which triggered this function
```
It's just `param` and not `params`

src/components/SRNEntry.vue
deepansh96: Agreed. We should just write what the function does and that's it

deepansh96 (changes requested):

src/components/SRNEntry.vue
```diff
@@ -104,72 +102,75 @@ export default {
...
         this.userIDList[0]["userID"] != ""
       );
     },
-    //checks if multiple entries are allowed and if one entry is already inputted
+    /** Whether multiple entries have been made by the user */
     ifUserEnteredMoreThanOne() {
       return !this.isSingleEntryOnly && this.userIdListLength > 1;
     },
-    /** for now, plio does not support multiple input entries */
+    /** For now, plio does not support multiple input entries */
```
deepansh96: I don't think this comment goes with the rule `docstrings should tell what the method does/performs`
I think you need to add that as well. The docstring you've added right now is just additional context information, we can keep that in as well.

src/components/SRNEntry.vue
```diff
@@ -104,72 +102,75 @@ export default {
...
     isSingleEntryOnly() {
       return this.redirectTo == "plio";
     },
-    /* returns submit button as disabled if any of the following is true:
-     - no SRN has been typed
-     - input is invalid
-     - SRN hasn't been completely typed */
+    /**
+     * Whether the submit button is disabled
+     * true if any of the following conditions are met:
```
```suggestion
     * Returns true if any of the following conditions are met:
```

src/components/SRNEntry.vue
```diff
@@ -104,72 +102,75 @@ export default {
...
       return this.userIDList.slice(-1)[0];
     },
   },
   methods: {
-    /* determines how the input box should look like depending on which box it is. 
-      - if the input box is the current one (i.e. being typed) and if the input is invalid (invalid input message is being displayed),
-        then, the box has a border of red
-      - if the input box is a past one (i.e. already typed), then the box is shown as unclickable 
-      @param {Number} - index - index of the input box */
+    /** Determines how the input box should look like depending on which input field it is.
```
```suggestion
    /** Determines how the input box should look.
```
Only this is enough no?

src/components/SRNEntry.vue
```diff
@@ -179,50 +180,51 @@ export default {
...
-    @param {Event} - e - event triggered when a character is typed.  */
+    /** Checks to see if the input character is a number. Makes use of ASCII values.
+     * @param {Object} e - event triggered when a character is typed
+     */
     isValidNumericEntry(e) {
       if (e.keyCode >= 48 && e.keyCode <= 57) return true;
       else e.preventDefault();
     },
-    // adds a new dict to the array - userIDList
+    /** Adds a new dict to the array - userIDList */
```
```suggestion
    /** Adds a new object to the userIDList array */
```

src/components/SRNEntry.vue
```diff
@@ -179,50 +180,51 @@ export default {
...
     addNewEmptyField() {
       this.userIDList.push({ userID: "", valid: false });
     },
-    /* removes an element in the array at a given index. 
-    @param {Number} - index - the index of the input box where - button is clicked. */
+    /** Removes an element in the array at a given index.
+     * @param {Number} index - the index of the input box where - button is clicked
```
```suggestion
     * @param {Number} index - the index of the input box where "-" button is clicked
```

src/components/SRNEntry.vue
```diff
@@ -179,50 +180,51 @@ export default {
...
+     */
     removeInputField(index) {
       this.userIDList.splice(index, 1);
     },
-    // sets the invalid input message to null (default)
+    /** Sets the invalid input message to null (default) */
```
```suggestion
    /** Resets the invalid input message */
```
It's not setting it to `null` to be precise, it's setting it to an empty string. Also, by not mentioning what it's setting the value to be, we save ourselves the trouble of changing the docstring again in the future if our default value changes from an empty string to, let's say, `null`

src/components/SRNEntry.vue
```diff
@@ -179,50 +180,51 @@ export default {
...
     resetInvalidInputMessage() {
       this.invalidInputMessage = "";
     },
-    // sets the invalid login message to null (default)
+    /** Sets the invalid login message to null (default) */
```
```suggestion
    /** Resets the invalid login message */
```

src/components/SRNEntry.vue
```diff
@@ -179,50 +180,51 @@ export default {
...
+    resetEntry(index) {
+      this.userIDList[index]["userID"] = "";
     },
-    /* this function is called whenever the + button is clicked. The most recent typed entry is authenticated against the database. 
-    - if it does not exist and it is the first time being authenticated, handleIncorrectEntry() is called. 
-    - if the entry exists or if it is being authenticated the second time, the 'valid' flag of the entry is set to whatever the value is returned by the backend. 
-      Also, isCurrentUserValid and validateCount are reset to their default values, so that a possible next entry can be processed. */
+    /** This function is called whenever the + button is clicked.
+     * The most recent typed entry is authenticated against the database.
+     */
```
deepansh96: Maybe something on these lines - 
Authenticates the most recently typed entry against the database

src/components/SRNEntry.vue
```diff
@@ -179,50 +180,51 @@ export default {
...
     async addField() {
-      const latestUserID = parseInt(this.getLatestEntry["userID"]);
+      const latestUserID = parseInt(this.latestEntry["userID"]);
       if (!isNaN(latestUserID)) {
         await this.authenticateSRN(latestUserID);
-        if(!this.isCurrentUserValid && this.validateCount == 1){
+        if (!this.isCurrentUserValid && this.validateCount == 1) {
           this.handleIncorrectEntry(latestUserID);
         }
         if (this.isCurrentUserValid || this.validateCount > 1) {
```
deepansh96, line 231: Not able to comment on the actual code so commenting here

Taking a look at this if block, notice the third line
`this.resetValidFlag`
Is it a method? If it's a method, there's no parenthese to invoke it?

src/components/SRNEntry.vue
```diff
@@ -233,32 +235,21 @@ export default {
...
-        validateCount = 1
-        isCurrentUserValid = false
-      The state of the variables contradict the fact that 'A' is already authenticated. This means any of the two cases:
-      - A is valid, so validateCount = 0 and isCurrentUserValid = true
-      - A is invalid, so validateCount = 2 and isCurrentUserValid = false
-      So when the user clicks submit, 'A' is authenticated again, which is a behaviour we do not want. So to bypass this, validateCount is set to 2. 
-      (To better understand how this will help, understanding processForm() will help.) 
-      isCurrentUserValid flag is ignored because the value is already set in each user's valid flag. 
-      @params {Number} - index - index of input field to be removed */
+    /** This method is called whenever - button is clicked, to remove an input field
```
deepansh96: Maybe something on the lines - 
Removes the selected entry from the entry list and resets appropriate variables

src/components/SRNEntry.vue
```diff
@@ -233,32 +235,21 @@ export default {
...
+     * @param {Number} index - index of input field to be removed
+     */
     removeField(index) {
       this.resetInvalidInputMessage;
```
deepansh96: Again, if these are methods then why are they not invoked?
Or is it a feature of vue that for methods with no arguments, we don't have to write `()`?

src/components/SRNEntry.vue
```diff
@@ -233,32 +235,21 @@ export default {
...
     },
-    /*  this function is called whenever something is entered in the input box. It checks if the required number of characters are being typed. 
-    Until then, it prompts the user to type the required characters. 
-    If the user types more than the desired number of characters, the input is sliced and the user wont be able to see the extra characters.
-    @params {Event} - event - the event which triggered this function
-    @params {Number} - index - the index of the input field */
+
+    /** This function is called whenever something is entered in the input box.
+     * It checks if the required number of characters are being typed.
+     * @param {Event} event - the event which triggered this function
```
```suggestion
     * @param {Object} event - the event which triggered this function
```
There's no data type as `Event` in javascript

src/components/SRNEntry.vue
```diff
@@ -273,64 +264,52 @@ export default {
...
-        this.resetEntry(this.userIdListLength - 1)
+      if (this.invalidLoginMessage != "") {
+        this.resetEntry(this.userIdListLength - 1);
       }
     },
-    /* this method is called after the user clicks the submit button. Since all the ID's but the last are authenticated, this method calls authenticateSRN(). 
-    If the user is invalid and is being authenticated the first time, handleIncorrectEntry() is called. 
-    If the user is valid or is being authenticated the second time, the user is reidrected to their destination and an SQS message is sent. */
+
+    /** This method is called after the user clicks the submit button.
```
deepansh96: This still doesn't tell what the method does, it only says when this method is invoked.

src/components/SRNEntry.vue
```diff
@@ -273,64 +264,52 @@ export default {
...
+     */
```
```suggestion
    /** This method is called after the user clicks the submit button.*/
```

src/services/validation.js
```diff
@@ -1,17 +1,15 @@
...
+* firebaseAPI service is used to validate and it returns a boolean value, indicating whether the user is valid or not.
+* @param {String} userID - current ID being authenticated 
+* @param {Number} validateCount - indicates how many times the user has been validated 
+*/
+
+export async function validateSRN(userID, validateCount){
     var invalidLoginMessage = ""
 
-    isCurrentUserValid = await firebaseAPI.checkUserExists(userID);
+    var isCurrentUserValid = await firebaseAPI.checkUserExists(userID);
```
deepansh96:
```suggestion
    let isCurrentUserValid = await firebaseAPI.checkUserExists(userID);
```
We don't need this variable to be scoped outside of this method, so no need of `var`

deepansh96 (changes requested):

src/components/SRNEntry.vue
sudheshna-donthineni: Yeah, you don't need the brackets if you don't have any arguments. Actually, it doesn't matter even if you have arguments, the whole event object gets sent, so arguments are included. Only when you have default values to those arguments, then the () make a difference.

src/components/SRNEntry.vue
```diff
@@ -217,8 +218,8 @@ export default {
     resetEntry(index) {
       this.userIDList[index]["userID"] = "";
     },
-    /** This function is called whenever the + button is clicked.
-     * The most recent typed entry is authenticated against the database.
+    /** This function is called whenever the "+"" button is clicked.
```
```suggestion
    /** This function is called whenever the "+" button is clicked.
```

deepansh96: Usually it helps a lot doing a self review before sending the PR for review

src/components/SRNEntry.vue
```diff
@@ -235,7 +236,8 @@ export default {
         }
       }
     },
-    /** This method is called whenever - button is clicked, to remove an input field
+    /** This method is called whenever - button is clicked.
```
```suggestion
    /** This method is called whenever "-" button is clicked.
```

deepansh96: Also it is super helpful if a reviewer makes a suggestion, one should look for places in the entire code where this suggestion could be applied.
Because I might've seen that this `minus` is not in quotes in one place, but I might not have checked other places. That is the job of the person who created the PR

deepansh96 (changes requested):

src/components/SRNEntry.vue
deepansh96: Need to fix this and all the other places where you're trying to invoke a method but the method is not invoked because of missing brackets

deepansh96 (approved):

PR #29, opened 2021-12-14
Program support

### LINKED ISSUE #34: Program support
- Various programs have different data and different UI.
- Build support in the app to be able to use it across programs

sudheshna-donthineni:
This code contains logic that displays the component `IdEntry` based on the program using the auth layer. 
- The URL will now contain `program` and `authType`.  
- Based on this program name, `getProgramData` API is called to retrieve all details regarding the program. 
- Basic validation login is based on the validation type (`basicValidationMapping`)
- For advanced validation, the database and the column name is also a variable now, based on the program.

The backend PR lies here https://github.com/avantifellows/auth-layer-backend/pull/4


src/components/IdEntry.vue
```diff
@@ -15,7 +15,7 @@
   >
     <!-- title -->
     <p class="text-xl sm:text-2xl md:text-2xl lg:text-3xl xl:text-4xl mx-auto font-bold">
-      Enter your SRN / अपना SRN दर्ज करें
+      {{ this.textObject["displayText"] }}
```
pritamps: Need a better name here. `displayText ` tells us nothing about what this is for.  Also (it has to be changed anyway, but notion and code should be consistent), according to the notion JSON, shouldn't this be just `display`?

src/pages/Home.vue
```diff
@@ -1,21 +1,28 @@
...
+      :program="program"
+      :authType="authType"
     />
   </div>
 </template>
 
 <script>
-import SRNEntry from "@/components/SRNEntry.vue";
+import IdEntry from "@/components/IdEntry.vue";
+import firebaseAPI from "@/services/API/getProgramData.js";
```
dalmia: Wouldn't it be better to use a more functionality-specific name that is independent of what we are using on the backend?
```suggestion
import ProgramAPIService from "@/services/API/getProgramData.js";
```

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
       default: "",
       type: String,
     },
+    program: {
+      /** The program the user falls under. Eg: HaryanaStudents, DelhiStudents */
```
dalmia: the convention is that it should be above the param, just as we write docstrings above functions

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+      default: "",
+      type: String,
+    },
+    authType: {
+      /** The authentication method used by the user */
```
dalmia: same comment on docstring as above.

deepansh96:
A few points --
- I saw that there was no issue created and linked to this PR. We've written all the main, important steps in our development lifecycle doc [here](https://www.notion.so/avantifellows/Development-Lifecycle-9acaf86e4fce4c3081b089e1b6482349). Please make an issue and link here! It helps a lot in the long run

- Also, for this feature, PRs had to be made in two repos, `auth-layer` and `auth-layer-backend`. The best practice in these cases is to keep the names of the PRs identical. Right now, this PR is called `Different programs support` and the one in the `auth-layer-backend` repo is called `Program support`. Ideally both names should be the same. Again, this helps a lot in the long term. For us (when we forget about the code) and for others who will work on the project.

- Also, when two PRs are interlinked and depend on each other, one needs to mention that in the description.  (more about it in the doc linked above)

- These last two points were not written in the doc so I've added them. I'll keep adding things from my side regarding the process. Please make sure you go through the doc again

src/components/IdEntry.vue
deepansh96: No need to write `this.` in the template part of a vue file. The template already has access to the component instance (which is referred to by `this`). 

Screenshot 2021-12-16 at 12 56 48 PM

In the docs above it tells us that `this` is required to access the component instance **only in lifecycle hooks**.

Also see this https://v3.vuejs.org/guide/template-syntax.html#text

src/components/IdEntry.vue
```diff
@@ -80,39 +79,55 @@
...
 import { redirectToDestination } from "@/services/redirectToDestination.js";
 import { sendSQSMessage } from "@/services/API/sqs";
-
-const NUMBER_OF_INPUTS_ALLOWED = 10;
+import { typeToFunctionMap } from "@/services/basicValidationMapping.js";
 const authType = "SRN";
 
 export default {
-  name: "SRNEntry",
+  name: "IdEntry",
```
deepansh96: Why name the component `IdEntry.vue`? I feel like just `Entry.vue` should be descriptive enough!

src/router/index.js
```diff
@@ -10,7 +10,9 @@ const routes = [
           purpose: route.query.purpose,
           purposeParams: route.query.subPurpose,
           redirectTo: route.query.redirectTo,
-          redirectID: route.query.redirectID
+          redirectID: route.query.redirectID,
+          program: route.query.program,
```
dalmia: this seems like something that should be a part of purposeParams. with purposeParams being like a dictionary.

src/services/API/checkUser.js
```diff
@@ -5,11 +5,15 @@ export default {
     /**
       * Validates that the ID exists
       * @param {String} inputUserID - the id that needs to be validated
+      * @param {String} collectionName - firestore collection against which the ID needs to be validated
+      * @param {String} columnName - name of the column which contains the ID
```
dalmia: this is coupling backend and the frontend, right? can we decouple them please so that the only connection between them is the API endpoint?

src/services/validation.js
```diff
@@ -4,23 +4,21 @@ import firebaseAPI from "@/services/API/checkUser.js";
 * firebaseAPI service is used to validate and it returns a boolean value, indicating whether the user is valid or not.
 * @param {String} userID - current ID being authenticated
 * @param {Number} validateCount - indicates how many times the user has been validated
+* @param {String} collectionName - firestore collection against which the ID needs to be validated
+* @param {String} columnName - name of the column which contains the ID
```
dalmia: same point on decoupling backend and frontend as above

src/services/basicValidationMapping.js
```diff
@@ -0,0 +1,11 @@
+/** Mapping between basicValidationType and the basic validation login
+ * @param {Object} event - event triggered when a character is typed
+ */
```
dalmia: I think this was mentioned in a previous PR too, but there is a slight correction in how this is done!
```suggestion
/** 
  * Mapping between basicValidationType and the basic validation login
  * @param {Object} event - event triggered when a character is typed
  */
```

src/services/basicValidationMapping.js
```diff
@@ -0,0 +1,11 @@
...
+export const typeToFunctionMap = {
+    "numeric": function (event){
+        return (event.keyCode >= 48 && event.keyCode <= 57);
+    },
+    "alphanumeric": function(event){
```
dalmia: inconsistency in the spacing between `function` and `(event)` across the two functions.
I think this was mentioned in a previous PR too that it would be very helpful during reviews if all these basic issues are taken care of on the local system itself either through pre-commits or through a linter. Have you set them up?

src/services/basicValidationMapping.js
```diff
@@ -0,0 +1,11 @@
+/** Mapping between basicValidationType and the basic validation login
+ * @param {Object} event - event triggered when a character is typed
+ */
+export const typeToFunctionMap = {
```
dalmia: since this function will be imported in a different file for use, just having `type` would make it very hard to understand the type of what, no?

```suggestion
export const validationTypeToFunctionMap = {
```

src/services/API/sqs.js
```diff
@@ -20,7 +20,7 @@ const sqsClient = new SQSClient({
 * Everything, except the userIDList, is extracted from the auth layer URL
 */
 
-export async function sendSQSMessage(purpose, purposeParams, redirectTo, redirectID, userIDList, authType) {
+export async function sendSQSMessage(purpose, purposeParams, redirectTo, redirectID, userIDList, authType, programName) {
```
dalmia: as mentioned in a previous comment, programName feels like it should be a part of purposeParams

src/services/API/getProgramData.js
```diff
@@ -0,0 +1,22 @@
...
+    {
+
+        const params = {
+            program : programName
+        }
+        return new Promise( (resolve) => {
+            client.post(getProgramDataEndpoint, JSON.stringify(params)).then((response) => {
+                resolve(response.data)
+            })
+
```
dalmia:
```suggestion
```

src/services/API/getProgramData.js
```diff
@@ -0,0 +1,22 @@
...
+import { getProgramDataEndpoint } from "@/services/API/endpoints.js";
+
+export default {
+    /**
+      * Validates that the ID exists
+      * @param {String} programName - the name of the program whose data needs to be retrieved
+      */
+    getProgramData(programName)
+    {
+
```
dalmia:
```suggestion
```

src/services/API/getProgramData.js
```diff
@@ -0,0 +1,22 @@
+import {client} from "@/services/API/rootClient.js"
+import { getProgramDataEndpoint } from "@/services/API/endpoints.js";
+
+export default {
+    /**
+      * Validates that the ID exists
```
dalmia: the description seems to be wrong

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
   },
   computed: {
+    /** Whether multiple ID can be entered */
     isMultipleIDEntryAllowed() {
       return this.redirectTo == "plio";
     },
   },
+  async created() {
+    /** Program name is sent to the API to retrieve all details */
+    var programData = await firebaseAPI.getProgramData(this.program);
```
dalmia: why `var`?

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+
+    let textObject = programData["text"]["default"];
+    this.textObject["placeholderText"] = textObject["placeholder"];
+    this.textObject["displayText"] = textObject["display"];
+    this.textObject["addButtonText"] = textObject["addButton"];
+    this.textObject["invalidInputMessage"] = textObject["invalid"]["input"];
+    this.textObject["invalidLoginMessage"] = textObject["invalid"]["login"];
+    this.textObject["submitButtonText"] = textObject["submitButton"];
```
dalmia: why not?
```suggestion
    this.textObject = {
        placeholderText: textObject["placeholder"],
        ...
    }
```

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+
+    let inputObject = programData["input"];
+    this.inputObject["basicInputValidationType"] = inputObject["basicValidationType"];
+    this.inputObject["maxLengthOfId"] = inputObject["maxLengthOfId"];
+    this.inputObject["mode"] = inputObject["mode"];
+    this.inputObject["type"] = inputObject["type"];
+
+    let dataSourceObject = programData["dataSource"];
+    this.dataSourceObject["name"] = dataSourceObject["name"];
+    this.dataSourceObject["column"] = dataSourceObject["column"];
```
dalmia: same suggestion as above.

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+    return {
+      textObject: {
+        placeholderText: "",
+        displayText: "",
+        addButtonText: "",
+        invalidInputMessage: "",
+        invalidLoginMessage: "",
+        submitButtonText: "",
+      },
+      maxNumberOfInput: 0,
```
dalmia: comment for what this means? it's not clear to me what is the purpose of this.

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+      inputObject: {
+        basicInputValidationType: "",
+        maxLengthOfId: 0,
+        mode: null,
+        type: null,
+      },
```
dalmia: suggesting a different name for the variable + that of inner params to avoid redundancy.

```suggestion
      inputParams: {
        basicValidationType: "",
        maxLength: 0,
        mode: null,
        type: null,
      },
```

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+      dataSourceObject: {
```
dalmia:
```suggestion
      dataSourceParams: {
```

src/pages/Home.vue
dalmia: can you give some examples?

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
       default: "",
       type: String,
     },
+    program: {
```
dalmia, line 61: same suggestion as before of including it under purposeParams

src/components/IdEntry.vue
```diff
@@ -80,39 +79,55 @@
...
   },
   computed: {
+    /** Returns the input mode stored against the program */
+    inputMode() {
+      return this.inputObject["mode"];
+    },
+    /** Returns the input type stored against the program */
+    inputType() {
+      return this.inputObject["type"];
+    },
```
dalmia: can you give some examples? what is the difference between input type and input mode?

src/components/IdEntry.vue
```diff
@@ -56,7 +56,7 @@
     <span class="mx-auto text-red-700 text-base mb-1" v-if="isInvalidLoginMessageShown">{{
       invalidLoginMessage
     }}</span>
-    <!-- add srn button -->
+    <!-- add button -->
```
dalmia: please be a bit more descriptive with the comments so that it is immediately clear what something is doing. 

```suggestion
    <!-- button to add another input -->
```

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+    },
+    authType: {
+      /** The authentication method used by the user */
+      default: "",
+      type: String,
+    },
+  },
+  data() {
+    return {
+      textObject: {
```
dalmia:
```suggestion
      textParams: {
```

src/pages/Home.vue
dalmia: understood now after reading everything end-to-end what it means. a better name could be `maxInputsAllowed` or `maxEntriesAllowed`

src/components/IdEntry.vue
```diff
@@ -295,14 +315,21 @@ export default {
...
+        this.validateCount,
+        this.dataSourceObject["name"],
+        this.dataSourceObject["column"]
+      );
       this.isCurrentUserValid = userValidationResponse.isCurrentUserValid;
       this.validateCount = userValidationResponse.validateCount;
-      this.invalidLoginMessage = userValidationResponse.invalidLoginMessage;
       this.isLoading = false;
 
+      if (this.validateCount == 1) {
```
dalmia: i thought we had decided to wrap such conditions under methods/computed properties with meaningful names, right?

src/components/IdEntry.vue
dalmia: can we use a better name?
like `Input.vue` or `InputForm.vue`?

src/components/IdEntry.vue
```diff
@@ -137,9 +152,9 @@ export default {
     /**
      * Whether the submit button is disabled
      * Returns true if any of the following conditions are met:
-     * - no SRN has been typed
+     * - no ID has been typed
```
dalmia:
```suggestion
     * - no ID has been entered
```

src/components/IdEntry.vue
```diff
@@ -137,9 +152,9 @@ export default {
...
      * - input is invalid
-     * - SRN hasn't been completely typed
+     * - ID hasn't been completely typed
```
dalmia:
```suggestion
     * - ID hasn't been completely entered
```

src/components/IdEntry.vue
```diff
@@ -191,12 +206,13 @@ export default {
...
      * @param {Object} event - event triggered when a character is typed
      */
-    isValidNumericEntry(event) {
-      if (event.keyCode >= 48 && event.keyCode <= 57) return true;
-      else event.preventDefault();
+    isValidEntry(event) {
+      if (typeToFunctionMap[this.inputObject["basicInputValidationType"]](event)) {
+        return true;
+      } else event.preventDefault();
     },
```
dalmia: need to handle the case when the validation type is not present in the mapping

src/components/IdEntry.vue
```diff
@@ -80,39 +79,55 @@
...
     redirectTo: String,
     redirectID: String,
     purpose: String,
     purposeParams: String,
+    textObject: Object,
+    inputObject: Object,
+    dataSourceObject: Object,
+    maxNumberOfInput: Number,
+    program: String,
+    authType: String,
```
dalmia: suggestions as before on renaming all of these and restructuring some

dalmia (changes requested):

src/services/validation.js
```diff
@@ -4,23 +4,21 @@ import firebaseAPI from "@/services/API/checkUser.js";
...
-    let isCurrentUserValid = await firebaseAPI.checkUserExists(userID);
+export async function validateID(userID, validateCount, collectionName, columnName){
+    let isCurrentUserValid = await firebaseAPI.checkUserExists(userID, collectionName, columnName);
     if(isCurrentUserValid){
         return {
             isCurrentUserValid: isCurrentUserValid,
             validateCount: validateCount,
-            invalidLoginMessage: invalidLoginMessage
             }
         }
```
deepansh96: Some of these lines are not linted. Do you have a linter set up on your VSCode for plain js files?
Even if that's not set up, you should set up pre-commit for this repo as well, locally AND as a github actions.

Set it up once and forget about linting issues forever!

src/services/validation.js
```diff
@@ -4,23 +4,21 @@ import firebaseAPI from "@/services/API/checkUser.js";
...
 * @param {Number} validateCount - indicates how many times the user has been validated
+* @param {String} collectionName - firestore collection against which the ID needs to be validated
+* @param {String} columnName - name of the column which contains the ID
 */
 
-export async function validateSRN(userID, validateCount){
-    let invalidLoginMessage = ""
-
-    let isCurrentUserValid = await firebaseAPI.checkUserExists(userID);
+export async function validateID(userID, validateCount, collectionName, columnName){
```
deepansh96: We'll have to think this through thoda. I think we were also supposed to support different **_types_** of datastores right?

In the JSON that you've written in the doc, I also see a `type` option. (attaching a screenshot)
Screenshot 2021-12-16 at 7 41 02 PM

But I don't see any code which is checking the type and then redirecting to different methods? Althought right now we only have one type but it's not being checked anywhere. 

Ideally this file, `validation.js` should deal with all kinds of validations from all kinds of data sources. The method called `validateID` should take in the `userID`, `validateCount` and **_all the datastore details_**. 

When we're just passing `collectionName` and `columnName`, we're coupling this validation method to a firebase type datastore. I'm guessing you would've thought of that already. We'll deal with other kinds of datastore when they come but we should start building in the support! Like, in our case currently, pass the whole datastore object, check it's type, and call the `firebaseAPI.checkUserExists()` method if the type is firebase.

We'll discuss more on the call!

src/services/API/checkUser.js
deepansh96: Responding to Aman's comment on this line - 
I'm not sure we have any other option? `checkUserExists` is a function deployed on cloud which needs these two pieces of information to work. What would you suggest we do?

src/pages/Home.vue
deepansh96: Responding to Aman's comment - 
Yes agreed.
Two things to notice here
- A file is called `getProgramData`. There's a "get" in front of the name. That usually signifies a getter or a method. A file should be named by what it contains. This file contains a method that retrieves program data using an API call. So maybe we should think about renaming this file. Just program data is also fine! Because right now it contains code to GET the program data. Maybe we might decide to PUSH/UPDATE/DELETE some program data in the future. Then it won't make sense to create separate files for that code right? All of that code can come under a file called `ProgramData.js`.

- And as Aman suggested, the name should be changed. Again, the fact that it's a **_firebase_** API is something that can be changed in the future. Also, there's no need for this component (Home.vue) to know what kind of cloud provider or backend are we talking to. It just needs to know what a particular method will do.

--------

- Just sharing below what we did in Plio and why we follow that pattern everywhere. It made our lives so much easier/
- In Plio, we would have lots of different types of actions being taken on a Question object in the backend. So we made a file called `Question.js`. In that file we wrote something like
```js
export default {
    getQuestion() { ... },
    deleteQuestion() { ... },
    updateQuestion() { ... }
}
```
- What this code above is doing is exporting an _**object**_ that _**contains**_ these three methods. It's not exporting the methods directly. There's a difference. So while importing this in the parent file, we would write something like `import QuestionAPIService from "../../Question.js"`
- This `QuestionAPIService` is an _**object**_ whose keys are those methods you can see in the code above. And that is why to access a method, we would have to write `QuestionAPIService.updateQuestion(...)`

src/pages/Home.vue
deepansh96: @sudheshna-donthineni I think we're gonna ditch all the other params right after this program thing is defined?

We talked about this on the call no? A little confused now because then how would we know the destination information? Oh maybe through the session thing pritam is working on?

src/pages/Home.vue
```diff
@@ -38,11 +45,69 @@ export default {
...
+    /** Whether multiple ID can be entered */
     isMultipleIDEntryAllowed() {
       return this.redirectTo == "plio";
     },
   },
+  async created() {
+    /** Program name is sent to the API to retrieve all details */
+    var programData = await firebaseAPI.getProgramData(this.program);
+
+    let textObject = programData["text"]["default"];
```
deepansh96: What is this `default`? In the JSON doc on notion I could see no `default`?
image

src/pages/Home.vue
deepansh96: Responding to Aman's comment - 
I don't think there's need for that as well. 

@sudheshna-donthineni  So what you've done here is you got a programData object from the API call to firebase. After getting it, you're extracting things one by one and putting them into a local variable called textObject, and then passing it to the IdEntry component.

Why not just pass the whole object directly to the IdEntry component without all this parsing? 
If there's some null handling to be done, some edge cases to manage, you can do that here but this is just picking up from one place and putting it into another no?

Same with all the other objects below. All you would need to do is change some variables names in your IdEntry.vue

src/router/index.js
deepansh96: Responding to aman - we can't send dictionaries as query params no?

src/services/API/endpoints.js
```diff
@@ -1 +1,2 @@
 export const checkUserEndpoint = "/checkForUser"
+export const getProgramDataEndpoint = "/getProgramData"
```
deepansh96: Maybe not valid right now but will be soon, we have to think this over as well.

If other people can also use the auth layer, they might have a different place where they're storing their programData (or groupData) so we need to provide the support for hooking up their data store to get the program data file. @pritamps

src/services/basicValidationMapping.js
```diff
@@ -0,0 +1,11 @@
...
+    "numeric": function (event){
+        return (event.keyCode >= 48 && event.keyCode <= 57);
+    },
+    "alphanumeric": function(event){
+        return (event.keyCode >=48 && event.keyCode <= 57) || (event.keyCode >= 65 && event.keyCode <= 90) || (event.keyCode >= 97 && event.keyCode <= 122);
+    }
```
deepansh96: Just a little refactoring can be done here!
Here, for every key, you're directly linking the function to that key.

A better way to write this would be to define two functions separately above this `export const ...` line.
These functions can have names like `isNumeric(input)` or `isAlphaNumeric(input)`. And then when those functions are defined and have a proper name, you can link those functions to they keys `numeric` and `alphanumeric` in the object that you're exporting.

Oh btw you will also be able to use `isNumeric` inside `isAlphaNumeric` function directly without repeating code :P

deepansh96 (changes requested):

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,15 @@
+name: AF Auth Layer
```
pritamps, line 1: This should have been a separate PR @sudheshna-donthineni . Would have quickly gotten merged by itself.

src/components/Entry.vue
```diff
@@ -68,8 +68,7 @@
         ></inline-svg>
         <div class="border-l-2 border-gray-500 pl-3">
           <p class="leading-tight">
-            Add another SRN <br />
-            एक और SRN दर्ज करें
+            {{ programData["text"]["default"]["addButton"] }}
```
pritamps: Is there error checking that these params are present in the JSON?

src/components/Entry.vue
```diff
@@ -80,43 +79,60 @@
...
+      invalidLoginMessage: "",
       isLoading: false,
+      invalidInputMessage: null, // whether the input being entered by the user matches the basic validation criteria
     };
   },
+
   computed: {
+    /** Returns the input mode stored against the program */
+    inputMode() {
+      return this.programData.input.mode;
```
pritamps, line 119: 1. Add error checking to make sure all required fields exists
2. Add documentation to README explaining what the required fields are, and what are optional (if any)

src/components/Entry.vue
```diff
@@ -80,43 +79,60 @@
...
+    },
+
+    /** Returns the input type stored against the program */
+    inputType() {
+      return this.programData.input.type;
+    },
+
+    /** Returns the placeholder text stored against the program */
+    inputBoxPlaceholderText() {
+      return this.programData.text.default.placeholder;
```
pritamps, line 129: 1. Add error checking to make sure all required fields exist in the JSON
2. Add documentation to README explaining what the required fields are, and what are optional (if any)

src/components/Entry.vue
```diff
@@ -158,31 +178,70 @@ export default {
...
+      return this.programData.input.basicValidationType;
+    },
+
+    /** Returns the invalid input message stored against each program */
+    invalidInputText() {
+      return this.programData.text.default.invalid.input;
+    },
+
+    /** Returns the maximum number of ID's a user can enter */
+    maxNumberOfInput() {
```
pritamps: Why not name the method to `maxNumberOfIds` -- both comment and variable name seem to indicate this

src/components/Entry.vue
```diff
@@ -313,7 +385,7 @@ export default {
      */
     async authenticate() {
       let latestUserID = parseInt(this.latestEntry["userID"]);
-      await this.authenticateSRN(latestUserID);
+      await this.authenticateID(latestUserID);
```
pritamps, line 397: There seem to be a lot of common code between `Entry.vue` and `OTP.vue`, no?

pritamps (changes requested):

.github/workflows/deploy_to_s3_staging.yml
```diff
@@ -40,7 +40,10 @@ jobs:
         VUE_APP_AWS_ACCESS_KEY_ID: ${{ secrets.VUE_APP_AWS_SQS_ACCESS_KEY }}
         VUE_APP_AWS_SECRET_KEY: ${{ secrets.VUE_APP_AWS_SQS_SECRET_ACCESS_KEY }}
         VUE_APP_BASE_URL_MEET: 'https://meet.google.com/'
-
+        VUE_APP_AWS_REGION: ${{ secrets.VUE_APP_AWS_REGION }}
+        VUE_APP_AWS_SQS_URL: ${{ secrets.VUE_APP_AWS_SQS_URL }}
+        VUE_APP_BASE_URL_YOUTUBE: ${{ secrets.VUE_APP_BASE_URL_YOUTUBE }}
```
deepansh96: Should this really be a secret? :o

src/components/Entry.vue
```diff
@@ -68,7 +68,7 @@
         ></inline-svg>
         <div class="border-l-2 border-gray-500 pl-3">
           <p class="leading-tight">
-            {{ this.textObject["addButtonText"] }}
+            {{ programData["text"]["default"]["addButton"] }}
```
deepansh96: computed variable?

src/components/Entry.vue
```diff
@@ -79,7 +79,7 @@
       class="bg-primary hover:bg-primary-hover text-white font-bold shadow-xl uppercase text-lg mx-auto p-4 mt-4 rounded disabled:opacity-50 btn"
       :disabled="isSubmitButtonDisabled"
     >
-      {{ this.textObject["submitButtonText"] }}
+      {{ submitButtonDisplayText }}
```
deepansh96, line 82: computed variable?

src/components/Entry.vue
```diff
@@ -88,50 +88,51 @@
...
     redirectTo: String,
     redirectID: String,
     purpose: String,
     purposeParams: String,
-    textObject: Object,
-    inputObject: Object,
-    dataSourceObject: Object,
-    maxNumberOfInput: Number,
+    programData: Object,
     program: String,
```
deepansh96, line 101: programName?

src/components/Entry.vue
```diff
@@ -88,50 +88,51 @@
...
   },
   data() {
     return {
       userIDList: [{ userID: "", valid: false }], //array containing user-ids and a valid flag for each
-      invalidInputMessage: null, // whether the input is in correct format
       isCurrentUserValid: false, // whether the current user is valid
       validateCount: 0, // count the number of times the user has been validated
       invalidLoginMessage: "",
       isLoading: false,
+      invalidInputMessage: null, // whether the input being entered by the user matches the basic validation criteria
```
deepansh96: The comment seems a little off
As it's a message, the comment should be something like "the message to show when the input being entered by the user does not match ..."

src/components/Entry.vue
```diff
@@ -210,7 +254,11 @@ export default {
      * @param {Object} event - event triggered when a character is typed
      */
     isValidEntry(event) {
-      if (typeToFunctionMap[this.inputObject["basicInputValidationType"]](event)) {
+      if (
+        validationTypeToFunctionMap[this.programData["input"]["basicValidationType"]](
```
deepansh96: you forgot to use the basicValidationType computer property here

src/components/Entry.vue
```diff
@@ -280,17 +328,14 @@ export default {
...
         this.resetInvalidLoginMessage();
       } else {
         this.resetInvalidInputMessage();
       }
-      if (event.target.value.length > this.inputObject["maxLengthOfId"]) {
-        event.target.value = event.target.value.slice(
-          0,
-          this.inputObject["maxLengthOfId"]
-        );
+      if (event.target.value.length > this.maxLengthOfId) {
```
deepansh96: Also can you add this to the above if/else block?

src/pages/Home.vue
```diff
@@ -1,113 +1,77 @@
...
       type: String,
     },
+    /** Subcategory of the purpose. Eg: plio -> means the attendance is for a plio link */
     purposeParams: {
-      /** Subcategory of the purpose. Eg: plio -> means the attendance is for a plio link */
       default: "",
       type: String,
     },
+    /** The program the user falls under. Eg: HaryanaStudents, DelhiStudents */
     program: {
```
deepansh96, line 61: programName?

src/services/basicValidationMapping.js
```diff
@@ -1,11 +1,20 @@
-/** Mapping between basicValidationType and the basic validation login
+/**
```
deepansh96: I guess all this will get updated when you merge the OTP support PR into this right?

deepansh96 (changes requested):

src/components/Entry.vue
deepansh96: Yess the ideal scenario will be to have just one component which will render things based on the JSON

src/services/basicValidationMapping.js
sudheshna-donthineni: Yeah.

src/pages/Home.vue
sudheshna-donthineni: Decided, no? This will be the last step after merging the OTP and youtube PR.

.github/workflows/pre-commit.yml
sudheshna-donthineni: Sorry!

pritamps (approved):
This looks good to me! But wait for @deepansh96 to approve also!

PR #30, opened 2021-12-15
Youtube support

### LINKED ISSUE #32: Need to support YouTube as destination

sudheshna-donthineni:
Added YouTube as a destination. So far this format works for both YouTube live as well as a standard YouTube video. I don't think any AF YouTube class link is made, so couldn't test with those.

Sample test links:
https://www.youtube.com/watch?v=LH38LvRdgPA - YouTube live
https://www.youtube.com/watch?v=pvws6jJsH9U - YouTube video


pritamps (changes requested):
Can you add an example to the README of how this will work. Otherwise this PR looks great! :)

PR #31, opened 2021-12-23
OTP authentication

### LINKED ISSUE #33: Adding OTP authentication

sudheshna-donthineni:
- Added a component `OTP.vue`, which is rendered whenever `authType` is `OTP`.
- The JSON looks like
```
"HaryanaTeachers": {
        "text": {
            "default": {
                "display": "Enter your phone number / अपना फोन नंबर दर्ज करें",
                "placeholder": "Your phone number / आपका फोन नंबर",
                "submitButton":"SUBMIT / जमा करें",
                "invalid": {
                    "input": "Please type 10 numbers / कृपया १० संख्या टाइप करें"
                    },
                "enterOTP": "Your OTP / अपना OTP",
                "requestOTP": "Request OTP / अनुरोध OTP"
            },
            "en": {
                "display": "Enter your phone number",
                "placeholder": "Your phone number",
                "submitButton":"SUBMIT",
                "invalid": {
                    "input": "Please type 10 numbers"
                    },
                "enterOTP": "Your OTP",
                "requestOTP": "Request OTP"
            },
            "hi": {
                "display": "अपना फोन नंबर दर्ज करें",
                "placeholder": "आपका फोन नंबर",
                "submitButton":"जमा करें",
                "invalid": {
                    "input": "कृपया १० संख्या टाइप करें"
                    },
                "enterOTP": "अपना OTP",
                "requestOTP": "अनुरोध OTP"
            }

            },
        "input": {
            "mode": "numeric",
            "type": "tel",
            "basicValidationType": "numeric",
            "maxLengthOfId": 10
            },
        "dataSource": {
            "type": "Firestore",
            "name": "HaryanaTeachers",
            "url": "https://console.firebase.google.com/u/1/project/avantifellows/firestore/data/~2FHaryanaTeachers~2F0",
            "column": "mobile_number"
            }
        }
```
- If you see `validation.js`, there are two separate methods for checking the ID vs checking a phone number. The reason for this is, for a phone number, we don't need `validateCount`. It is only one try since the validation is only based on us checking if the phone number exists in our database or not. It does not have to do anything with the number being entered wrong.


sudheshna-donthineni:
Something I wanted to ask was if you look at the `requestOTP` method, the only status code being referred to is 200. There is no other user error code. If the response status is something other than 200, it's usually an internal server error or something. What message could we write? I've seen websites do "Oh no! Something bad happened!".

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+  </div>
+
+  <!-- main div -->
+  <div
+    class="flex flex-col my-auto h-full py-32 space-y-6"
+    :class="{ 'opacity-20 pointer-events-none': isLoading }"
+  >
+    <!-- title -->
+    <p class="text-xl sm:text-2xl md:text-2xl lg:text-3xl xl:text-4xl mx-auto font-bold">
+      {{ programData["text"]["default"]["display"] }}
```
pritamps: So basically right now, the backend supports multiple languages, and the frontend doesn't (since you're always picking the default)?

Can you add a task to Kanban about enabling localization and with some info on the current status?

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+    </p>
+
+    <!--text box to enter phone number-->
+    <div
+      class="flex flex-row justify-center"
+      v-for="(input, index) in userIDList"
+      :key="`idInput-${index}`"
+    >
+      <div>
```
pritamps: This `div` seems to be indented wrong? Shoudl be same as the previous div no?

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+  computed: {
+    /** Returns the input mode stored against the program */
+    inputMode() {
+      return this.programData["input"]["mode"];
+    },
+    /** Returns the input type stored against the program */
+    inputType() {
+      return this.programData["input"]["type"];
+    },
+    /** Returns the placeholder text stored against the program */
```
pritamps: What is the placeholder for? Just expand the comment to say where the placeholder is shown, and also change method name

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+        this.userIDList[0]["userID"].length < this.programData["input"]["maxLengthOfId"]
+      );
+    },
+    /** Whether input being typed is in the correct format */
+    isInvalidInputMessageShown() {
+      return this.invalidInputMessage != null;
+    },
+  },
+  methods: {
+    /** Determines how the input box should look.
```
pritamps, line 309: Expand comment on what the conditions are that are used to make this decision?

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+     * @param {Number} index - index of the input box
+     */
+    calculateInputClasses() {
```
pritamps: `calculate` seems weird since it implies a calculation. Maybe `selectInputBoxClasses`?

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+        )
+      ) {
+        return true;
+      } else event.preventDefault();
+    },
+    /** Resets the invalid input message */
+    resetInvalidInputMessage() {
+      this.invalidInputMessage = "";
+    },
+    /** This function is called whenever something is entered in the input box.
```
pritamps, line 340: For all methods -- leave an empty line between two functions

src/components/OTP.vue
```diff
@@ -0,0 +1,292 @@
...
+    /** Function that verifies the OTP */
+    async verifyOTP() {
+      const response = await OTPAuth.verifyOTP(
+        parseInt(this.userIDList[0]["userID"]),
+        parseInt(this.OTPCode)
+      );
+      this.displayOTPMessage = mapVerifyStatusCodeToMessage[response.status.toString()];
+      if (response.status == 200) {
+        this.authenticate();
+      }
```
pritamps, line 390: There's no `else` here? WHat happens on wrong OTP entry?

src/services/API/otp.js
```diff
@@ -0,0 +1,34 @@
...
+            phone: phoneNumber
+        }
+        return new Promise( (resolve) => {
+            OTPClient.post(sendOTPEndpoint, null, {params}).then((response) => {
+                resolve(response)
+            })
+
+        })
+    },
+     /*  Verifies OTP entered by user */
```
pritamps: Empty line between methods :)

src/services/OTPCodes.js
```diff
@@ -0,0 +1,12 @@
...
+    311: "This OTP does not exist / यह OTP मौजूद नहीं है",
+    309: "You have exceeded maximum number of attempts. Please try after sometime / आप प्रयासों की अधिकतम संख्या को पार कर चुके हैं। कृपया कुछ देर बाद प्रयास करें",
+    301: "OTP token is expired / OTP टोकन की समय सीमा समाप्त हो गई है",
+    310: "This OTP is incorrect / यह OTP गलत है",
+    308: "You are re-trying too early. Please wait for sometime / आप बहुत जल्दी पुनः प्रयास कर रहे हैं। कृपया कुछ देर प्रतीक्षा करें ",
+  }
+
+  export function mapSendStatusCodeToMessage(code) {
+    return code == 200 ? "OTP has been sent! / OTP भेज दिया गया है!" : "Oh no!"
+
```
pritamps: What about including other status codes in the map?

pritamps (changes requested):
I'm not so sure about including the full URL in the data source, especially since we don't have any security enabled? Is there a way to add API key auth to firebase also?

src/components/OTP.vue
sudheshna-donthineni: That's right!
The task is already added.

src/components/OTP.vue
sudheshna-donthineni: So, only when the status is 200, they are redirected. Otherwise, the `displayOTPMessage` is updated with the error message.

src/services/OTPCodes.js
sudheshna-donthineni: So, I left a comment in the PR. There are no user status codes. There are status codes like 500 or 404, which are more our side errors. 
We see errors like "oh our bad" types. That's what I'm thinking we'll need to add.

pritamps:
> Something I wanted to ask was if you look at the `requestOTP` method, the only status code being referred to is 200. There is no other user error code. If the response status is something other than 200, it's usually an internal server error or something. What message could we write? I've seen websites do "Oh no! Something bad happened!".

Just say "Unable to send OTP. Try again?"

.env.example
```diff
@@ -5,3 +5,5 @@ VUE_APP_AWS_SQS_SECRET_KEY=''
 VUE_APP_BASE_URL_MEET=''
 VUE_APP_AWS_REGION=''
 VUE_APP_AWS_SQS_URL=''
+VUE_APP_OTP_URL=''
+VUE_APP_TEACHER_WEB_APP_URL=''
```
deepansh96: You'll remove this no? As this will be passed in the query params itself

.env.example
```diff
@@ -5,3 +5,5 @@ VUE_APP_AWS_SQS_SECRET_KEY=''
 VUE_APP_BASE_URL_MEET=''
 VUE_APP_AWS_REGION=''
 VUE_APP_AWS_SQS_URL=''
+VUE_APP_OTP_URL=''
```
deepansh96:
```suggestion
VUE_APP_OTP_SERVICE=''
```

docs/ENV.MD
```diff
@@ -21,3 +21,9 @@ AWS region
 
 ## VUE_APP_AWS_SQS_URL
 Base URL for AWS SQS
+
+## VUE_APP_OTP_URL
```
deepansh96:
```suggestion
## VUE_APP_OTP_SERVICE
```

docs/ENV.MD
```diff
@@ -21,3 +21,9 @@ AWS region
...
+Base URL for OTP service
+
+## VUE_APP_TEACHER_WEB_APP_URL
+URL for Teacher Web App
```
deepansh96: Will be removed based on the comment above

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+  </div>
+
+  <!-- main div -->
+  <div
+    class="flex flex-col my-auto h-full py-32 space-y-6"
+    :class="{ 'opacity-20 pointer-events-none': isLoading }"
+  >
+    <!-- title -->
+    <p class="text-xl sm:text-2xl md:text-2xl lg:text-3xl xl:text-4xl mx-auto font-bold">
+      {{ programData["text"]["default"]["display"] }}
```
deepansh96: Won't it be easier to read if we turn it into a computed variable?

deepansh96: Also, in javascript you can simply do this
`programData.text.default.display`
but make sure you handle the undefined check. If some key doesn't exist, it should not throw an error

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+    </p>
+
+    <!--text box to enter phone number-->
+    <div
+      class="flex flex-row justify-center"
+      v-for="(input, index) in userIDList"
```
deepansh96:
```suggestion
      v-for="(phoneNumber, index) in phoneNumberList"
```
How does this look?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+      <input
+        v-model="input.userID"
+        :type="inputType"
+        :inputmode="inputMode"
+        pattern="[0-9]*"
+        :placeholder="placeholderText"
+        required
+        @keypress="isValidEntry($event)"
+        class="border-2 rounded-sm p-4 mx-auto border-gray-500 focus:border-gray-800 focus:outline-none"
+        :class="selectInputBoxClasses"
```
deepansh96, line 39: Maybe add the `class="border-2 rounded-sm p-4 mx-auto border-gray-500 focus:border-gray-800 focus:outline-none"` in the `selectInputBoxClasses` variable itself?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+
+    <!-- invalid input message  -->
+    <span class="mx-auto text-red-700 text-base mb-1" v-if="isInvalidInputMessageShown">{{
+      invalidInputMessage
+    }}</span>
+
+    <!--text box to enter OTP-->
+    <div class="flex flex-col justify-center" v-if="isOTPSent">
+      <p class="text-md sm:text-l md:text-l lg:text-xl xl:text-2xl mx-auto font-bold">
+        {{ programData["text"]["default"]["enterOTP"] }}
```
deepansh96: computed variable + undefined/null checks

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+    </div>
+
+    <!-- button to request for OTP -->
+    <button
+      @click="sendOTP"
+      class="bg-primary hover:bg-primary-hover text-white uppercase text-lg mx-auto p-4 mt-4 rounded-3xl border-2 disabled:opacity-50"
+      :disabled="isRequestOTPButtonDisabled"
+      v-show="!isOTPSent"
+    >
+      {{ programData["text"]["default"]["requestOTP"] }}
```
deepansh96: computed variable + undefined/null checks

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+    }}</span>
+
+    <!-- submit button -->
+    <button
+      v-show="isOTPSent"
+      @click="verifyOTP"
+      class="bg-primary hover:bg-primary-hover text-white font-bold shadow-xl uppercase text-lg mx-auto p-4 mt-4 rounded disabled:opacity-50 btn"
+      :disabled="isSubmitButtonDisabled"
+    >
+      {{ programData["text"]["default"]["submitButton"] }}
```
deepansh96: computed variable + undefined/null checks

deepansh96:
I got this warning in the console when I visited this URL - 
```
localhost:8080/?purpose=attendance&redirectTo=teacher-web-app&redirectID=https://haryana-teachers-af.web.app/&subPurpose=liveclass&program=HaryanaTeachers&authType=OTP
```

image

deepansh96:
Can you add some padding to the left and right? because when the width goes below 400 something, the text just sticks to the left part. I think it'll look neater if there was some padding around

https://user-images.githubusercontent.com/20169829/147929367-06eb614e-2aaf-457d-acd1-a8a48c20f096.mov

deepansh96:
Also, why two different styles for the buttons? The send otp button's style is different than the submit button's style

https://user-images.githubusercontent.com/20169829/147929700-368fa70c-ac69-441e-9a56-da9dc331119d.mov

deepansh96:
I get two types of warnings

1. Whenever I'm typing my mobile number into the textbox, a warning related to `placeholderText` comes up

2. When I've entered all of my 10 digits, and if I try to add more by typing more letters, a warning related to `userID` is shown

https://user-images.githubusercontent.com/20169829/147930180-f6645c02-1366-48c3-80eb-d907a0d96f95.mov

deepansh96:
Suggestions 
### For the programData/groupData JSON
- Instead of "Your OTP", if we use "Enter your OTP" that'll make more sense right?
- Also, in hindi it's written "apna OTP". I guess the correct translation is "aapka OTP" 

### Visual
- The `OTP has been sent` message is in red color. Red color signifies something went wrong, I feel :p 
- When I enter my OTP, it takes a very small time to redirect and during that time nothing happens on the screen. I feel if we can put the app once again in a loading state once a user enters the OTP then it'll make for a good experience. Also we can avoid bugs like pressing the submit button again

src/services/validation.js
```diff
@@ -33,9 +43,25 @@ async function checkUserInFirestore(userID, validateCount, collectionName, colum
     }
 
 }
-export async function validateID(userID, validateCount, collectionName, columnName, dataSourceType){
-    if(dataSourceType == "Firestore")
+
+/** This function checks the data source to see which database API needs to be called.
+* @param {String} userID - current ID being authenticated
+* @param {Number} validateCount - indicates how many times the user has been validated
+* @param {String} collectionName - firestore collection against which the ID needs to be validated
```
deepansh96: This will not include "firestore" anymore right?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+      this.isLoading = true;
+      this.userIDList[0]["valid"] = await validateID(
+        this.userId,
+        this.validateCount,
+        this.programData["dataSource"]["name"],
+        this.programData["dataSource"]["column"],
+        this.authType
+      );
+      this.isLoading = false;
+    },
```
deepansh96: In this method, you're sending the `name` and `column` to the `validateID` method separately. Why not send the whole `this.programData['datasource']` object and let `validateID` take what it needs?

I'm saying this because right now we have firestore, so all you need for firestore is a `collectionName` and `columnName`. But if the case of some other type of datastore, we might need some more details. So instead of restricting the `validateID` function to only receive `collectionName` and `columnName`, we should ideally pass the whole datasource object and let the `validateID` function handle what to pick up

src/services/redirectToDestination.js
```diff
@@ -8,16 +8,17 @@ import { sendSQSMessage } from "@/services/API/sqs";
 * @param {String} authType - extracted from auth layer URL
 */
 
-export function redirectToDestination(purposeParams, userIDList, redirectID, redirectTo, authType){
+export function redirectToDestination(purposeParams, userIDList, redirectID, redirectTo, authType, program){
```
deepansh96: There's a typo in the docstring for this method. Sorry had to comment here cause I could not comment on the line above.

src/services/redirectToDestination.js
```diff
@@ -8,16 +8,17 @@ import { sendSQSMessage } from "@/services/API/sqs";
...
     let redirectURL = "";
     let fullURL = "";
+    let queryParams = "";
+    var userID = userIDList[0]["userID"]
 
     switch(redirectTo){
         case 'plio':
-            var userID = userIDList[0]["userID"]
             redirectURL = process.env.VUE_APP_BASE_URL_PLIO;
             var url = new URL(redirectURL + redirectID);
```
deepansh96:
```suggestion
            let url = new URL(redirectURL + redirectID);
```

src/services/redirectToDestination.js
```diff
@@ -8,16 +8,17 @@ import { sendSQSMessage } from "@/services/API/sqs";
 * @param {String} authType - extracted from auth layer URL
 */
 
-export function redirectToDestination(purposeParams, userIDList, redirectID, redirectTo, authType){
+export function redirectToDestination(purposeParams, userIDList, redirectID, redirectTo, authType, program){
     let redirectURL = "";
     let fullURL = "";
+    let queryParams = "";
+    var userID = userIDList[0]["userID"]
```
deepansh96:
```suggestion
    let userID = userIDList[0]["userID"]
```

src/services/redirectToDestination.js
```diff
@@ -33,6 +34,14 @@ export function redirectToDestination(purposeParams, userIDList, redirectID, red
             fullURL = redirectID
             break;
 
+        case 'teacher-web-app':
+            queryParams = new URLSearchParams({
+                teacherID: userID,
+                programID: program
```
deepansh96: This is not actually `programID` right? This is the program name. can we change it?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+
+export default {
+  name: "OTP",
+  props: {
+    redirectTo: String,
+    redirectID: String,
+    purpose: String,
+    purposeParams: String,
+    programData: Object,
+    program: String,
```
deepansh96, line 128: Should be programName no?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+    authType: String,
+  },
+  data() {
+    return {
+      userIDList: [{ userID: "", valid: false }],
```
deepansh96: Thinking if this should be renamed to `phoneNumberList` or not, what do you think?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+      isUserValid: false,
+      isOTPSent: false,
+      OTPCode: "",
+      isLoading: false,
+      displayOTPMessage: "",
+      invalidInputMessage: null,
+    };
```
deepansh96: Comments please! Even one liner comments are also good

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * Whether the submit button is disabled
+     * Returns true if any of the following conditions are met:
+     * - if 'Request OTP' button is enabled
+     * - if OTP is being typed
+     */
+    isSubmitButtonDisabled() {
+      return !(this.isRequestOTPButtonDisabled || this.OTPCode.length > 0);
+    },
```
deepansh96: Here, you wrote `is OTP is being typed` which is not super clear to me at least.
I'm guessing that the submit button will stay disabled if one hasn't typed the OTP string right?

Maybe update the docstring and the code to something like this
```js
/**
     * Whether the submit button is disabled
     * - if 'Request OTP' button is enabled
     * - if OTP hasn't been typed yet
     */
    isSubmitButtonDisabled() {
      return !this.isRequestOTPButtonDisabled || this.OTPCode.length == 0
    },
```
If you see above, the docstring exactly matches what's written in the code

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+
+    /** Checks if the current input entry has the required number of characters */
+    isCurrentEntryIncomplete() {
+      return this.userId.length < this.programData["input"]["maxLengthOfId"];
```
deepansh96: Maybe you can save `this.programData["input"]["maxLengthOfId"]` as a computed variable so you can reuse this value elsewhere as well.

Also, it's fine to write this as well `this.programData.input.maxLengthOfId`. Makes the code more readable
Also in that computed variable, it's best to have one default value as well. If by chance, the keys `input` or `maxLengthOfId` is not present, it should not throw an error and stop the execution. A good default would be 10 here!

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+    /** Whether input being typed is in the correct format */
+    isInvalidInputMessageShown() {
+      return this.invalidInputMessage != null;
+    },
+  },
+  methods: {
+    /** Determines how the input box should look.
+     * - If an input error needs to be displayed, the box has a a red border.
+     * - Otherwise, it has an opacity of 30.
+     * @param {Number} index - index of the input box
```
deepansh96: I don't see any param here!

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+        },
+      ];
+    },
+
+    /** Calls the mapping function to validate the typed character
+     * @param {Object} event - event triggered when a character is typed
+     */
+    isValidEntry(event) {
+      if (
+        validationTypeToFunctionMap[this.programData["input"]["basicValidationType"]](
```
deepansh96: Would be best to make all these values inside `this.programData` as computed variables. Would reduce the code a lot and we can reuse them elsewhere easily

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+
+    /** Resets the invalid input message */
+    resetInvalidInputMessage() {
+      this.invalidInputMessage = "";
+    },
+
+    /** This function is called whenever something is entered in the input box.
+     * It checks if the required number of characters are being typed.
+     * @param {Object} event - the event which triggered this function
+     * @param {Number} index - the index of the input field
```
deepansh96: I dont see this param passed here

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+      if (event.target.value.length == 0) {
+        this.invalidInputMessage = "";
+      } else if (event.target.value.length < this.programData["input"]["maxLengthOfId"]) {
+        this.invalidInputMessage = this.programData["text"]["default"]["invalid"][
+          "input"
+        ];
+      } else {
+        this.resetInvalidInputMessage();
+      }
+      if (event.target.value.length > this.programData["input"]["maxLengthOfId"]) {
```
deepansh96: Why not add this condition as another else if attached to the above if/else logic?

src/components/OTP.vue
```diff
@@ -0,0 +1,308 @@
...
+      );
+      this.displayOTPMessage = mapVerifyStatusCodeToMessage[response.status.toString()];
+      if (response.status == 200) {
+        this.authenticate();
+      }
+    },
+
+    /** This method authenticates the phone number.
+     */
+    async authenticatePhoneNumber() {
```
deepansh96, line 395: I got confused between `authenticate()` method and `authenticatePhoneNumber` method
Maybe rename `authenticate` to `authenticateAndRedirect`?

src/pages/Home.vue
```diff
@@ -11,16 +11,30 @@
...
   </div>
+  <!-- OTP component -->
+  <div v-else-if="isAuthTypeOTP && programData">
+    <OTP
+      :redirectTo="redirectTo"
+      :redirectID="redirectID"
+      :purpose="purpose"
+      :purposeParams="purposeParams"
+      :programData="programData"
+      :program="program"
```
deepansh96, line 22: programName would be better?

src/pages/Home.vue
```diff
@@ -68,8 +82,13 @@ export default {
     isAuthTypeID() {
       return this.authType == "ID";
     },
+    /** Whether authentication method chosen is OTP */
+    isAuthTypeOTP() {
+      return this.authType == "OTP";
+    },
   },
   async created() {
+    console.log(this.program, this.authType);
```
deepansh96: I guess you forgot to remove this

src/services/API/otp.js
```diff
@@ -0,0 +1,34 @@
+import {OTPClient} from "@/services/API/rootClient.js"
+import { sendOTPEndpoint, verifyOTPEndpoint } from "@/services/API/endpoints.js";
```
deepansh96: pre commit is not working on js files?
on line 1 there's no space between the brackets and the `OTPClient` string, but on line 2 there is space present.
Can you please check if the pre commit is working on js files or not. Beacuse I think it's working fine on .vue files

src/services/API/otp.js
deepansh96:
```suggestion
     /**  Verifies OTP entered by user */
```

src/services/OTPCodes.js
```diff
@@ -0,0 +1,12 @@
...
+    311: "This OTP does not exist / यह OTP मौजूद नहीं है",
+    309: "You have exceeded maximum number of attempts. Please try after sometime / आप प्रयासों की अधिकतम संख्या को पार कर चुके हैं। कृपया कुछ देर बाद प्रयास करें",
+    301: "OTP token is expired / OTP टोकन की समय सीमा समाप्त हो गई है",
+    310: "This OTP is incorrect / यह OTP गलत है",
+    308: "You are re-trying too early. Please wait for sometime / आप बहुत जल्दी पुनः प्रयास कर रहे हैं। कृपया कुछ देर प्रतीक्षा करें ",
+  }
+
+  export function mapSendStatusCodeToMessage(code) {
+    return code == 200 ? "OTP has been sent! / OTP भेज दिया गया है!" : "Unable to send OTP. Try again?"
+
```
deepansh96: comments please

src/services/redirectToDestination.js
```diff
@@ -8,16 +8,17 @@ import { sendSQSMessage } from "@/services/API/sqs";
 * @param {String} authType - extracted from auth layer URL
 */
 
-export function redirectToDestination(purposeParams, userIDList, redirectID, redirectTo, authType){
+export function redirectToDestination(purposeParams, userIDList, redirectID, redirectTo, authType, program){
     let redirectURL = "";
     let fullURL = "";
+    let queryParams = "";
```
deepansh96:
```suggestion
    let finalURLQueryParams = "";
```

Better readability

src/services/validation.js
```diff
@@ -1,14 +1,24 @@
 import firebaseAPI from "@/services/API/checkUser.js";
 
-/** This function validates an entry against the database (for this usecase, it is firestore).
+/** This function validates a phone number against Firestore.
+* firebaseAPI service is used to validate and it returns a boolean value, indicating whether the user is valid or not.
+* @param {String} phoneNumber - phone number being authenticated
```
deepansh96:
```suggestion
* @param {String} phoneNumber - phone number being validated
```

src/services/validation.js
```diff
@@ -1,14 +1,24 @@
...
+* @param {String} columnName - name of the column which contains the ID
+*/
+async function checkPhoneNumberInFirestore(phoneNumber, collectionName, columnName){
+    return await firebaseAPI.checkUserExists(phoneNumber, collectionName, columnName);
+}
+
+
+/** This function validates an entry against Firestore.
 * firebaseAPI service is used to validate and it returns a boolean value, indicating whether the user is valid or not.
 * @param {String} userID - current ID being authenticated
```
deepansh96:
```suggestion
* @param {String} userID - current ID being validated
```

src/services/validation.js
```diff
@@ -33,9 +43,25 @@ async function checkUserInFirestore(userID, validateCount, collectionName, colum
...
 
 }
-export async function validateID(userID, validateCount, collectionName, columnName, dataSourceType){
-    if(dataSourceType == "Firestore")
+
+/** This function checks the data source to see which database API needs to be called.
+* @param {String} userID - current ID being authenticated
+* @param {Number} validateCount - indicates how many times the user has been validated
+* @param {String} collectionName - firestore collection against which the ID needs to be validated
+* @param {String} columnName - name of the column which contains the ID
```
deepansh96: as mentioned in a previous comment, why not send the whole datasource to this method instead of sending columnName and collectionName separately?

src/services/validation.js
```diff
@@ -33,9 +43,25 @@ async function checkUserInFirestore(userID, validateCount, collectionName, colum
...
+        if(authType == "ID")
     {
-        return checkUserInFirestore(userID, validateCount, collectionName, columnName)
+        return checkUserIdInFirestore(userID, validateCount, collectionName, columnName)
     }
+    else if(authType == "OTP")
+    {
+        return checkPhoneNumberInFirestore(userID, collectionName, columnName)
+    }
+}
```
deepansh96: If you see, this part has incorrect indentation! I guess you need to debug why pre commit is not working

deepansh96 (changes requested):

src/services/validation.js
sudheshna-donthineni: I didn't understand this comment.

src/services/redirectToDestination.js
sudheshna-donthineni: did you mean the purposeParams?

.github/workflows/deploy_to_s3_staging.yml
```diff
@@ -40,10 +40,10 @@ jobs:
         VUE_APP_AWS_ACCESS_KEY_ID: ${{ secrets.VUE_APP_AWS_SQS_ACCESS_KEY }}
         VUE_APP_AWS_SECRET_KEY: ${{ secrets.VUE_APP_AWS_SQS_SECRET_ACCESS_KEY }}
         VUE_APP_BASE_URL_MEET: 'https://meet.google.com/'
-        VUE_APP_AWS_REGION: ${{ secrets.VUE_APP_AWS_REGION }}
```
deepansh96, line 43: Any reason for removing region?

src/components/OTP.vue
```diff
@@ -15,39 +15,45 @@
     :class="{ 'opacity-20 pointer-events-none': isLoading }"
   >
     <!-- title -->
-    <p class="text-xl sm:text-2xl md:text-2xl lg:text-3xl xl:text-4xl mx-auto font-bold">
-      {{ programData["text"]["default"]["display"] }}
+    <p
+      class="text-xl sm:text-2xl md:text-2xl lg:text-3xl xl:text-4xl mx-auto font-bold px-6"
+    >
+      {{ inputBoxDisplayText }}
```
deepansh96: This is actually title right?
Would make more sense if the variable name includes title

src/components/OTP.vue
```diff
@@ -15,39 +15,45 @@
...
         pattern="[0-9]*"
-        :placeholder="placeholderText"
+        :placeholder="inputBoxPlaceholderText"
         required
-        @keypress="isValidEntry($event)"
+        @keypress="isValidPhoneNumber($event)"
         class="border-2 rounded-sm p-4 mx-auto border-gray-500 focus:border-gray-800 focus:outline-none"
         :class="selectInputBoxClasses"
-        @input="updateUserId($event)"
+        @input="updatephoneNumber($event)"
```
deepansh96: If you're changing the method name here from `updateUserId` to `updatePhoneNumber`, it means you're renaming userId to phoneNumber. If you're taking that decision then all instances of userId should be renamed to phoneNumber, like for example, `phoneNumberObject.userID` will be renamed to `phoneNumberObject.phoneNumber`

src/components/OTP.vue
```diff
@@ -106,45 +112,45 @@ export default {
...
   data() {
     return {
-      userIDList: [{ userID: "", valid: false }],
-      isUserValid: false,
-      isOTPSent: false,
-      OTPCode: "",
+      phoneNumberList: [{ userID: "", valid: false }], // contains the phone number entered by the user
+      isUserValid: false, // whether the user exists in our database
+      isOTPSent: false, // has the OTP been sent
+      OTPCode: "", // string that contain the OTP code entered by the user
```
deepansh96:
```suggestion
      OTPCode: "", // string that contains the OTP code entered by the user
```

src/components/OTP.vue
```diff
@@ -106,45 +112,45 @@ export default {
...
     inputBoxPlaceholderText() {
-      return this.programData["text"]["default"]["placeholder"];
+      return this.programData.text.default.placeholder;
     },
 
-    /** Checks if any userID has been entered */
-    isAnyUserIDPresent() {
-      return this.userIDList != undefined && this.userId != "";
+    /** Checks if any phoneNumber has been entered */
+    isAnyphoneNumberPresent() {
```
deepansh96:
```suggestion
    isAnyPhoneNumberPresent() {
```

src/components/OTP.vue
```diff
@@ -154,46 +160,75 @@ export default {
...
-      return this.userId.length < this.programData["input"]["maxLengthOfId"];
+      return this.phoneNumber.length < this.maxLengthOfId;
     },
 
     /** Whether input being typed is in the correct format */
     isInvalidInputMessageShown() {
       return this.invalidInputMessage != null;
     },
+
+    /** Returns the heading text for the input box */
```
deepansh96, line 200:
```suggestion
    /** Returns the title text for the input box */
```

src/components/OTP.vue
```diff
@@ -154,46 +160,75 @@ export default {
...
+    inputBoxDisplayText() {
+      return this.programData.text.default.display;
+    },
+
+    /** Returns the text for the OTP code input box */
```
deepansh96:
```suggestion
    /** Returns the title text for the OTP code input box */
```

src/components/OTP.vue
```diff
@@ -154,46 +160,75 @@ export default {
...
+    enterOTPInputBoxDisplayText() {
+      return this.programData.text.default.enterOTP;
+    },
+
+    /** Returns the text for the button to request an OTP */
```
deepansh96:
```suggestion
    /** Returns the text for the Request OTP button */
```

src/components/OTP.vue
```diff
@@ -154,46 +160,75 @@ export default {
...
+    requestOTPButtonDisplayText() {
+      return this.programData.text.default.requestOTP;
+    },
+
+    /** Returns the text for the submit button */
+    submitButtonDisplayText() {
+      return this.programData.text.default.submitButton;
+    },
+
+    /** Returns the maximum length of the ID */
```
deepansh96, line 220:
```suggestion
    /** Returns the maximum allowed length of the ID */
```

src/components/OTP.vue
```diff
@@ -154,46 +160,75 @@ export default {
...
+    maxLengthOfId() {
```
deepansh96: Should be `maxLengthOfPhoneNumber` ideally

src/components/OTP.vue
```diff
@@ -154,46 +160,75 @@ export default {
...
-        {
-          "border-red-600 focus:border-red-600": this.invalidInputMessage,
-          "pointer-events-none opacity-30": !this.invalidInputMessage,
-        },
-      ];
+      let baseStyle =
+        "border-2 rounded-sm p-4 mx-auto border-gray-500 focus:border-gray-800 focus:outline-none";
+      return this.invalidInputMessage
+        ? baseStyle + "border-red-600 focus:border-red-600"
+        : baseStyle + "pointer-events-none opacity-30";
```
deepansh96: Why this syntax? In Vue docs this syntax is suggested
```js
return [
    {
        "border-red-600 focus:border-red-600": this.invalidInputMessage,
        "pointer-events-none opacity-30": !this.invalidInputMessage,
    },
    "border-2 rounded-sm p-4 mx-auto border-gray-500 focus:border-gray-800 focus:outline-none"
]
```

src/services/basicValidationMapping.js
```diff
@@ -1,20 +1,26 @@
-/**
- * Mapping between basicValidationType and the basic validation login
+/** Function to check if the key press by user is a number.
```
deepansh96:
```suggestion
/** Function to check if the key pressed by user is a number.
```

src/services/basicValidationMapping.js
```diff
@@ -1,20 +1,26 @@
...
-
-function isNumeric(event){
-    return (event.keyCode >= 48 && event.keyCode <= 57);
+function isNumeric(event) {
+  return event.keyCode >= 48 && event.keyCode <= 57;
 }
 
-function isAlphanumeric(event){
-    return isNumeric(event) || (event.keyCode >= 65 && event.keyCode <= 90) || (event.keyCode >= 97 && event.keyCode <= 122)
+/** Function to check if the key press by user is a number or an alphabet.
```
deepansh96:
```suggestion
/** Function to check if the key pressed by user is a number or an alphabet.
```

src/services/basicValidationMapping.js
```diff
@@ -1,20 +1,26 @@
...
 }
-export const validationTypeToFunctionMap = {
-    "numeric": isNumeric,
-    "alphanumeric": isAlphanumeric,
-    undefined: function (){
-        return false
-    }
 
-}
+/** Mapping between basicValidationType and the basic validation logic */
```
deepansh96:
```suggestion
/** Mapping between basicValidationType and the validation method to use */
```

src/services/redirectToDestination.js
```diff
@@ -1,60 +1,67 @@
...
+    case "teacher-web-app":
+      finalURLQueryParams = new URLSearchParams({
+        teacherID: userID,
+        programName: program,
+      });
+      fullURL = redirectID + "?" + finalURLQueryParams;
+      break;
+
+    default:
+      var purpose = "Error";
```
deepansh96, line 58: as we discussed, wrap them in blocks

src/services/redirectToDestination.js
```diff
@@ -1,60 +1,67 @@
...
+) {
+  let redirectURL = "";
+  let fullURL = "";
+  let finalURLQueryParams = "";
+  let userID = userIDList[0]["userID"];
+
+  switch (redirectTo) {
+    case "plio":
+      redirectURL = process.env.VUE_APP_BASE_URL_PLIO;
+      var url = new URL(redirectURL + redirectID);
```
deepansh96: as we discussed, wrap them in blocks

deepansh96: as we discussed, wrap them in blocks

deepansh96: as we discussed, wrap them in blocks

deepansh96: as we discussed, wrap them in blocks

src/services/validation.js
```diff
@@ -1,67 +1,86 @@
...
+      isCurrentUserValid: isCurrentUserValid,
+      validateCount: validateCount,
+    };
+  }
 
+  if (validateCount == 0) {
+    validateCount += 1;
+  } else if (validateCount == 1) {
+    validateCount += 1;
+  }
```
deepansh96: Why not combine both if conditions?

src/services/validation.js
```diff
@@ -1,67 +1,86 @@
...
-        return checkUserIdInFirestore(userID, validateCount, collectionName, columnName)
-    }
-    else if(authType == "OTP")
-    {
-        return checkPhoneNumberInFirestore(userID, collectionName, columnName)
+ * @param {String} userID - current ID being validated
+ * @param {Number} validateCount - indicates how many times the user has been validated
+ * @param {String} dataSourceObject - contains information about where and how the data is stored.
+ * @param {String} authType - the authentication method the user has used
+ */
```
deepansh96: The order of these is not the same as the order of arguments in the method below

src/services/validation.js
```diff
@@ -1,67 +1,86 @@
...
+export async function validateID(
+  userID,
+  dataSourceObject,
```
deepansh96:
```suggestion
  dataSource,
```
Just `dataSource` should be fine

src/services/validation.js
```diff
@@ -1,67 +1,86 @@
...
+  authType,
+  validateCount = 0
+) {
+  if (dataSourceObject["type"] == "Firestore") {
+    if (authType == "ID") {
+      return checkUserIdInFirestore(
```
deepansh96, line 70: if you're returning here then no need of an else if below.

you can just use an if below instead of an else if

src/services/redirectToDestination.js
deepansh96: There's a type in this docstring text
```
/** This is function is called when a user has to be redirected to the destination. Depending on the destination, the destination URL is built.
```

src/components/OTP.vue
pritamps: I find the way she's written it clearer. It seems logical with the `if` condition on `this.invalidInputMessage`

src/components/OTP.vue
```diff
@@ -0,0 +1,337 @@
...
+    },
+
+    /** Calls the mapping function to validate the typed character
+     * @param {Object} event - event triggered when a character is typed
+     */
+    isValidPhoneNumber(event) {
+      if (validationTypeToFunctionMap[this.basicValidationType](event)) {
```
pritamps, line 330: Since this is the OTP component, do we need this `validationTypeToFunctionMap`? Don't we know that it will be a phone number?

src/components/OTP.vue
```diff
@@ -0,0 +1,337 @@
...
+    },
+
+    /** Function that calls the API to send an OTP
+     * If the status is 200, the OTP has been sent.
+     * Otherwise, retry
+     */
+    async sendOTP() {
+      const response = await OTPAuth.sendOTP(parseInt(this.phoneNumber));
+      if (response.status == 200) {
+        this.displayOTPMessage = mapSendStatusCodeToMessage(response.status.toString());
```
pritamps: This line is common to both, so it can be taken out of the `if`. And `this.OTPSent` can be set in one line

src/components/OTP.vue
```diff
@@ -0,0 +1,337 @@
...
+      this.invalidInputMessage = "";
+    },
+
+    /** This function is called whenever something is entered in the input box.
+     * It checks if the required number of characters are being typed.
+     * @param {Object} event - the event which triggered this function
+     */
+    updatephoneNumber(event) {
+      if (event.target.value.length == 0) {
+        this.invalidInputMessage = "";
```
pritamps: `inputMessage` refers to what here? I think since here we are dealing with only phone numbers, a lot of the variable names can be simplified, like @deepansh96 suggested

pritamps (changes requested):
This should have been smaller PRs

src/components/OTP.vue
sudheshna-donthineni: Yeah, since it's OTP, the `basicValidationType` will be `numeric`. Just wanted it to be consistent across components.

.github/workflows/deploy_to_s3_staging.yml
sudheshna-donthineni: it's being added by the program support PR.

src/components/OTP.vue
sudheshna-donthineni: It's the message that keeps being displayed until the user types the required number of characters. "Please type 10 digits."

src/components/OTP.vue
pritamps: If consistent across components and same/similar validation methods are being used, why not combine the common code into one place?

src/components/OTP.vue
pritamps: Point here was that variable names can be simplified, since this is the OTP component and will clearly deal only with phone numebrs and OTPs

src/components/OTP.vue
sudheshna-donthineni: The validation methods are in one place. I was talking about syntax consistency. As in, instead of hardcoding the value, referring to the mapping.

sudheshna-donthineni:
> I'm not so sure about including the full URL in the data source, especially since we don't have any security enabled? Is there a way to add API key auth to firebase also?

Created an issue, going to look into it.

PR #41, opened 2022-01-17
Added local storage to OTP authentication

### LINKED ISSUE #42: Enable local storage for OTP auth
- Store phone number in local storage, so the user won't have to ask for an OTP each time.

sudheshna-donthineni:
- Basically, if a user is logs in with their phone number for the first time, then it is set in the store's state. 
- Next time user tries to log in, the phone number is retrieved from the local storage and the user is redirecting to the destination.

### DIFF
--- a/src/components/OTP.vue
+++ b/src/components/OTP.vue
@@ -311,8 +311,11 @@ export default {
   },
   mounted() {
     /** If user already logged in, get from store and redirect to destination */
-    this.phoneNumberList["0"]["userID"] = this.$store.getters["getUserPhoneNumber"];
-    this.authenticateAndRedirect();
+    let phoneNumber = this.$store.getters.getUserPhoneNumber;
+    if (phoneNumber) {
+      this.phoneNumberList["0"]["userID"] = phoneNumber;
+      this.authenticateAndRedirect();
+    }
   },
   methods: {
     /** Determines how the input box should look.
--- a/src/components/OTP.vue
+++ b/src/components/OTP.vue
@@ -145,8 +145,8 @@ export default {
   },
 
   computed: {
-    /** Retrieve phone number stored in vuex */
     ...mapState({
+      /** Retrieve phone number stored in vuex */
       statePhoneNumber: (state) => state.phoneNumber,
     }),
 
--- a/src/components/OTP.vue
+++ b/src/components/OTP.vue
@@ -147,7 +147,7 @@ export default {
   computed: {
     ...mapState({
       /** Retrieve phone number stored in vuex */
-      statePhoneNumber: (state) => state.phoneNumber,
+      storePhoneNumber: (state) => state.phoneNumber,
     }),
 
     /** Extracts phone number from list */
@@ -316,8 +316,8 @@ export default {
   },
   mounted() {
     /** If user already logged in, get from store and redirect to destination */
-    if (this.statePhoneNumber) {
-      this.phoneNumberList[0]["userID"] = this.statePhoneNumber;
+    if (this.storePhoneNumber) {
+      this.phoneNumberList[0]["userID"] = this.storePhoneNumber;
       this.authenticateAndRedirect();
     }
   },
--- a/src/components/OTP.vue
+++ b/src/components/OTP.vue
@@ -114,7 +114,7 @@ import {
   mapVerifyStatusCodeToMessage,
   mapSendStatusCodeToMessage,
 } from "@/services/OTPCodes.js";
-
+import { mapState, mapActions } from "vuex";
 const RESEND_OTP_TIME_OUT = 60;
 
 export default {
@@ -145,6 +145,10 @@ export default {
   },
 
   computed: {
+    /** Retrieve phone number stored in vuex */
+    ...mapState({
+      statePhoneNumber: (state) => state.phoneNumber,
+    }),
     /** Extracts phone number from list */
     phoneNumber() {
       return this.phoneNumberList["0"]["userID"];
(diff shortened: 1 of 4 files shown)


src/store/index.js
```diff
@@ -0,0 +1,42 @@
...
+  actions: {
+    setPhoneNumber({ commit }, phoneNumber) {
+      commit("setPhoneNumber", phoneNumber);
+    },
+  },
+  mutations: {
+    setPhoneNumber(state, phoneNumber) {
+      state.phoneNumber = phoneNumber;
+    },
+  },
```
deepansh96, line 21: Also add an action and mutation here which "unsets" the phone number, removes it from the store and localstorage. This is a kind of a convention we follow, that whenever we add an action+mutation to "set" something, we also add another action+mutation to "unset" it.

src/store/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+//encrypt and decrypt the localStorage
+let localStorage = new SecureLS({ isCompression: true });
+
+export default createStore({
+  state: {
+    phoneNumber: null,
+  },
+  getters: {
+    getUserPhoneNumber: (state) => {
```
deepansh96: Two points
- Getters are used just like computed variables. If you need to compute some value on top of an existing data property, you would use computed variables in a normal .vue component. Same case here. If you want to compute some value on top of a state variable, then you should use a getter. So ideally, in this case, a getter is not really needed. You can safely remove it and just use state variable instead.


- getters are like the computed variables but for store. So you can assume this to be a computed variable. And if you remember, the naming of computed variables is a little different from methods. If you were to use a getter here in this case, you can safely use the name `userPhoneNumber` instead of `getUserPhoneNumber`

src/components/OTP.vue
```diff
@@ -309,6 +309,14 @@ export default {
     /** The user type is set as soon as component is created */
     this.userType = this.groupData.userType;
   },
+  mounted() {
+    /** If user already logged in, get from store and redirect to destination */
+    let phoneNumber = this.$store.getters.getUserPhoneNumber;
```
deepansh96: Instead of using store as `this.$store.getters...` you should use `mapState` and `mapGetters`
you can mix the store variables into your own list of data variables and computed variables of a vue component and then you can use it directly like `this.userPhoneNumber`

Read more here https://next.vuex.vuejs.org/guide/state.html#the-mapstate-helper

src/components/OTP.vue
```diff
@@ -384,6 +392,7 @@ export default {
       const responseStatusCode = responseStatusCodeAndMessage[1];
       if (responseStatusMessage.trim() === "success") {
         this.isLoading = true;
+        this.$store.dispatch("setPhoneNumber", this.phoneNumber);
```
deepansh96: Use `mapActions` for this which is another helper like mapState. you can take a look at plio's code to get an idea how to use this

deepansh96 (changes requested):

deepansh96:
Also, I was curious, what are some cases where we might need to remove the phone number from the local store

pritamps (changes requested):
Have nothing to add to what @deepansh96 has said, except for a naming consistency thing: some places it's `userPhoneNumber`, some places it's `phoneNumber`. Just pick one :)

pritamps:
> Also, I was curious, what are some cases where we might need to remove the phone number from the local store

There shouldn't be any cases where we remove phone number from local store. The auth layer (for now) is only responsible for logging people in, not logging people out. Maybe there is a future where auth layer can deal with the full session of the user, but it's not there yet, and we shouldn't design for it.

src/store/index.js
sudheshna-donthineni: Since there is no unsetting action being done, is it necessary to include the function?

deepansh96: Haan waise leave it right now
I was wondering that we might need to remove it from store when the user logs out but that functionality is not there right now. Leave it then

src/components/OTP.vue
```diff
@@ -309,6 +309,14 @@ export default {
     /** The user type is set as soon as component is created */
     this.userType = this.groupData.userType;
   },
+  mounted() {
+    /** If user already logged in, get from store and redirect to destination */
+    let phoneNumber = this.$store.getters.getUserPhoneNumber;
+    if (phoneNumber) {
+      this.phoneNumberList["0"]["userID"] = phoneNumber;
```
dalmia: any reason we are storing it as `this.phoneNumberList["0"]` instead of `this.phoneNumberList[0]`

sudheshna-donthineni: No reason, I started using it that way and now it persists.

deepansh96: It means the same thing right? 
We should use [0] instead of ["0"]!

src/components/OTP.vue
```diff
@@ -145,6 +145,11 @@ export default {
   },
 
   computed: {
+    /** Retrieve phone number stored in vuex */
+    ...mapState({
+      statePhoneNumber: (state) => state.phoneNumber,
+    }),
```
deepansh96:
```suggestion
    ...mapState({
      /** Retrieve phone number stored in vuex */
      statePhoneNumber: (state) => state.phoneNumber,
    }),
```
I think it would make sense to move the comment inside, just above a particular state value that you're mapping. Because when more state variables will come, then you might want to add a comment for each of them

src/components/OTP.vue
```diff
@@ -145,6 +145,11 @@ export default {
   },
 
   computed: {
+    /** Retrieve phone number stored in vuex */
+    ...mapState({
+      statePhoneNumber: (state) => state.phoneNumber,
```
deepansh96:
```suggestion
      phoneNumber: (state) => state.phoneNumber,
```
This is also fine and much simpler

deepansh96 (changes requested):

src/components/OTP.vue
sudheshna-donthineni: It conflicts to a computed property, no?

PR #43, opened 2022-01-21
Sentry integration

rathorevaibhav:
### Summary

1. Sentry setup
2. Update to `.env.example` for Sentry DSN
3. Minor cleanups in readme and docs


.env.example
```diff
@@ -12,3 +12,6 @@ VUE_APP_BASE_URL_YOUTUBE=''
 
 ### setting up OTP authentication
 VUE_APP_OTP_SERVICE=''
+
+### sentry settings
+VUE_APP_SENTRY_DSN=''
```
sudheshna-donthineni, line 17: Is this only for production? Or does it work in staging as well?

sudheshna-donthineni (approved):

.env.example
rathorevaibhav: If the value is not empty/null, this will work. Otherwise, it doesn't. If needed, we can just add an env variable to staging and set this value. It'll start working.

rathorevaibhav: @sudheshna-donthineni based on discussion with Pritam, I've added a check in `main.js` to enable Sentry logs only on production. Please have a look.

pritamps (approved):
This looks good to me, but @sudheshna-donthineni should approve as well

PR #49, opened 2022-03-09
Added Localeswitcher to the Homepage

### LINKED ISSUE #48: Adding LocalSwitcher before Homepage to choose a language

gurleen3:
Fixes #48 

## Summary 

Displayed a popup to let the user choose which language do they wish to continue with (Hindi/English).

## Todos
- [ ] Show only the Locale Switch page and then the homepage


src/pages/Home.vue
```diff
@@ -71,6 +78,8 @@ export default {
   data() {
     return {
       groupData: null,
+      toggle: true,
```
sudheshna-donthineni, line 81: You shouldn't be using a variable to display the language picker. Because this way, everytime the app is refreshed, the picker shows up. It should show up only when the user is coming on to the page for the first time. Refer to how it is done in Plio.

src/pages/Home.vue
```diff
@@ -83,6 +92,12 @@ export default {
       return this.authType == "OTP";
     },
   },
+  methods: {
+    setLocale(value) {
+      this.locale = value;
+      console.log(this.locale);
```
sudheshna-donthineni: console log statements shouldn't exist in production code.

sudheshna-donthineni (changes requested):

src/pages/Home.vue
gurleen3: okay

PR #50, opened 2022-04-04
Migrating from Webpack to Vite

### LINKED ISSUE #54: Migrating from Webpack to Vite

sudheshna-donthineni:
- This code is just the initial set-up of Vite.


docs/ENV.MD
```diff
@@ -1,29 +1,29 @@
 # Environment variables
 This guide explains all the available configurations in your `.env` file so that you can configure your environments accordingly.
 
-## VUE_APP_PLIO_AF_API_KEY
+## VITE_APP_PLIO_AF_API_KEY
```
pritamps, line 4: This change is not needed. It's still a Vue app, just using Vite

docs/INSTALLATION.MD
```diff
@@ -16,8 +16,8 @@ For installing, follow steps at https://github.com/awslabs/git-secrets#installin
 Follow the steps below to set up the repo for development.
 1. Clone the repository and change the working directory
     ```sh
-    git clone https://github.com/avantifellows/auth-layer.git
-    cd auth-layer
+    git clone https://github.com/avantifellows/Portal.git
+    cd Portal
```
pritamps, line 20: Why captital `P` in Portal?

docs/ENV.MD
sudheshna-donthineni: It's a Vue app but the bundler is Vite. Before it was Vue CLI, so to read the env variables they needed to start with VUE_APP_. Now for Vite to read them, they need to start with VITE_APP

deepansh96: Not necessarily
`VITE_` is default but you can change it
https://vitejs.dev/config/index.html#envprefix

Ab you have made the change so let it be

docs/INSTALLATION.MD
deepansh96: Ab toh we can't change repo name :p 
It'll be lot of work

package.json
```diff
@@ -3,16 +3,17 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "serve": "vue-cli-service serve",
-    "build": "vue-cli-service build",
-    "lint": "vue-cli-service lint",
-    "deploy-staging": "vue-cli-service build --mode staging && aws --profile af-s3-bot --region ap-south-1 s3 sync ./dist s3://af-auth-layer-staging --delete --acl public-read",
-    "deploy": "vue-cli-service build && aws --profile af-s3-bot --region ap-south-1 s3 sync ./dist s3://af-auth-layer --delete --acl public-read"
+    "dev": "vite",
+    "build": "vite build",
```
deepansh96, line 7: you should also add this here
`preview: "vite build && vite preview"`

deepansh96: This is to check if a production build looks OK in a local environment

package.json
```diff
@@ -3,16 +3,17 @@
...
+    "lint": "vite lint",
+    "deploy-staging": "vite build --mode staging && aws --profile af-s3-bot --region ap-south-1 s3 sync ./dist s3://af-auth-layer-staging --delete --acl public-read",
+    "deploy": "vite build && aws --profile af-s3-bot --region ap-south-1 s3 sync ./dist s3://af-auth-layer --delete --acl public-read"
   },
   "dependencies": {
     "@aws-sdk/client-sqs": "^3.30.0",
     "autoprefixer": "^9.8.6",
     "axios": "^0.21.1",
+    "buffer": "^6.0.3",
```
deepansh96, line 17: what is this?

package.json
```diff
@@ -26,13 +27,12 @@
     "vuex-persistedstate": "^4.1.0"
   },
   "devDependencies": {
-    "@vue/cli-plugin-babel": "~4.5.0",
```
deepansh96, line 29: is babel included by default in vite?

src/components/Entry.vue
```diff
@@ -110,6 +110,9 @@ export default {
       isLoading: false,
       invalidInputMessage: null, // message to show when the input being entered does not match the ID format,
       userType: "", // differentiates between different kinds of users
+      loadingSpinner: new URL('../assets/images/loading_spinner.svg', import.meta.url).href,
+      deleteSvg: new URL('../assets/images/remove_circle.svg', import.meta.url).href,
+      addSvg: new URL('../assets/images/add_circle.svg', import.meta.url).href
```
deepansh96: There's a better way to do this, which I also encountered in a project I just set up.

You can use glob imports provided by vite to setup asset loading in one file and use it everywhere. https://vitejs.dev/guide/features.html#glob-import

Create a utility `.js` file which will do the asset handling. For example, `assets.js`
This is how the contents of the file might look like
```js
// inside assets.js

export default function useAssets() {
  // reference https://github.com/vitejs/vite/issues/1265#issuecomment-885948466
  const allJpegs = import.meta.globEager('/src/assets/*.jpeg')
  const allPngs = import.meta.globEager('/src/assets/*.png')
  const allSvgs = import.meta.globEager('/src/assets/*.svg')
  
return {
    loadingSpinnerIcon: jpegs['/src/assets/loading_spinner.svg'].default,
    deleteIcon: jpegs['/src/assets/remove_circle.svg'].default,
    .
    .
    .
  }
}
```

And then, in whatever file you want to use these, you can just do this
```js
// inside Entry.vue

import useAssets from '../../assets.js'

const assets = useAssets()
// then you can directly use `assets.loadingSpinner` and that will contain your SVG
```

deepansh96: Also, if possible, try to maintain consistency in naming!
Like one svg is named `deleteSvg` but another one is named `loadingSpinner`.
If you're adding `svg` at the end of the name, make sure to do that for all the svgs

deepansh96 (changes requested):

docs/ENV.MD
sudheshna-donthineni: I've tried with `VUE_` as the prefix, the env variables weren't read.

package.json
sudheshna-donthineni: Basically, this NodeJS library does not exist in Vite, so you have to install it ourselves. 
https://github.com/aws/aws-sdk-js-v3/issues/3373

package.json
sudheshna-donthineni: We don't need Babel anymore. Vite uses esbuild instead.

src/components/Entry.vue
```diff
@@ -89,6 +89,9 @@ import { validateID } from "@/services/validation.js";
 import { redirectToDestination } from "@/services/redirectToDestination.js";
 import { sendSQSMessage } from "@/services/API/sqs";
 import { validationTypeToFunctionMap } from "@/services/basicValidationMapping.js";
+import useAssets from '../assets/assets.js'
```
deepansh96: @ syntax?

deepansh96 (changes requested):

.github/workflows/deploy_to_s3_prod.yml
```diff
@@ -34,18 +34,18 @@ jobs:
...
+        VITE_APP_AWS_ACCESS_KEY_ID: ${{ secrets.VITE_APP_AWS_SQS_ACCESS_KEY }}
+        VITE_APP_AWS_SECRET_KEY: ${{ secrets.VITE_APP_AWS_SQS_SECRET_ACCESS_KEY }}
+        VITE_APP_BASE_URL_MEET: 'https://meet.google.com/'
+        VITE_APP_OTP_SERVICE: ${{ secrets.VITE_APP_OTP_SERVICE }}
+        VITE_APP_AWS_REGION: ${{ secrets.VITE_APP_AWS_REGION }}
+        VITE_APP_AWS_SQS_URL: ${{ secrets.VITE_APP_AWS_SQS_URL }}
+        VITE_APP_BASE_URL_YOUTUBE: 'https://www.youtube.com/watch?v='
+        VITE_APP_CHECK_USER_ENDPOINT: '/checkForUser'
+        VITE_APP_GET_GROUP_ENDPOINT: '/getGroupData'
+        VITE_APP_CLOUD_FUNCTIONS_BASE_URL: "https://us-central1-avantifellows.cloudfunctions.net"
```
deepansh96, line 48: is this function still in us region?

src/components/NoClassMessage.vue
```diff
@@ -0,0 +1,26 @@
...
+  <div class="fixed w-full top-1/4 my-5 flex justify-center">
+    <div
+      class="bg-white w-11/12 sm:w-9/12 lg:w-7/12 p-4 sm:p-10 rounded-lg border border-black"
+    >
+      <img
+        :src="AFLogo"
+        class="w-32  m-auto"
+      >
+      <p class="text-center text-2xl sm:text-4xl py-4 sm:py-8">
+        There is no class now. <br>Please check the timetable.
```
deepansh96: Can we also write the hindi version below it?

src/components/NoClassMessage.vue
```diff
@@ -0,0 +1,26 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "NoClassMessage",
+  data(){
+    return {
+      AFLogo: new URL('/AF_favicon.png', import.meta.url).href
```
deepansh96: can you import this png using the `useAssets` that you created?

src/components/SessionEntry.vue
```diff
@@ -0,0 +1,124 @@
...
+      :authType="authType"
+      :sessionId="sessionId"
+    />
+  </div>
+</template>
+
+<script>
+import NoClassMessage from "./NoClassMessage.vue";
+import Entry from "./Entry.vue";
+import groupAPIService from "../services/API/groupData";
```
deepansh96: @ syntax

deepansh96 (changes requested):

.github/workflows/deploy_to_s3_prod.yml
sudheshna-donthineni: Yeah, since I created the cron job, we decided to let this be for now. Since we'll be moving to elixir anyway.

src/components/NoClassMessage.vue
sudheshna-donthineni: I'm still waiting on the ops team to give me a message in the form they want. This is just a placeholder until then. Won't be merging until I get the message.

deepansh96: got it

src/assets/assets.js
```diff
@@ -5,5 +5,6 @@ export default function useAssets() {
       allSvgs["/src/assets/images/loading_spinner.svg"].default,
     deleteSvg: allSvgs["/src/assets/images/remove_circle.svg"].default,
     addSvg: allSvgs["/src/assets/images/add_circle.svg"].default,
+    AFLogoSvg: allSvgs["/AF_favicon.png"].default,
```
deepansh96: this is a png, not an svg.
just like you have `allSvgs`, you should make a new const called `allPngs` and then access aflogopng from there

deepansh96 (changes requested):

PR #51, opened 2022-04-06
added session ID as param

### LINKED ISSUE #53: Adding sessions support to auth layer

sudheshna-donthineni:
This code extracts sessionId param from URL and retrieves the session details. Based on whether the session is enabled or not, either the session entry component is displayed or a message is displayed to the user. 
(This branch is off the previous PR branch.)


src/components/SessionEntry.vue
```diff
@@ -0,0 +1,19 @@
...
+     props: {
+    sessionData: Object,
+     },
+     computed: {
+
+     },
+     async created() {
+
+   }
+};
```
deepansh96: is linting not working? ;_;

src/pages/Home.vue
```diff
@@ -1,6 +1,10 @@
 <template>
+<!-- UserMessage component -->
+<div v-if="!sessionEnabled">
+  <UserMessage/>
+</div>
   <!-- Entry component -->
```
deepansh96, line 7: You can put everything below in a div saying `v-else`.
That way you won't have to write `sessionEnabled` in every v-if condition going forward

src/pages/Home.vue
```diff
@@ -1,6 +1,10 @@
 <template>
+<!-- UserMessage component -->
+<div v-if="!sessionEnabled">
+  <UserMessage/>
```
deepansh96: can you think of a better name?
Or if you wanna keep it UserMessage, then make the actual message that's being shown to the user as a prop. Then it actually becomes generic enough.

src/pages/Home.vue
```diff
@@ -67,10 +81,17 @@ export default {
...
+      default: "",
+      type: String
+    }
   },
   data() {
     return {
-      groupData: null,
+      groupData: null, // stores details about a group
+      sessionData: null, // stores details about a session
+      sessionEnabled: false // whether a session is enabled
```
deepansh96: can you also add something like `doesGroupDataExists` which is a boolean?
Currently in the code you're using `groupData` as a proxy to check if group data is present or not
Would be better to have separate boolean indicating if group data is present or not

src/pages/Home.vue
```diff
@@ -84,8 +105,13 @@ export default {
     },
   },
   async created() {
-    /** Group name is sent to the API to retrieve all details */
-    this.groupData = await groupAPIService.getGroupData(this.group);
+    /** If sessionId exists in route, then retrieve session details. Otherwise, fallback is using group. */
```
deepansh96, line 113:
```suggestion
    /** If sessionId exists in route, then retrieve session details. Otherwise, fallback to using group data. */
```

src/pages/Home.vue
```diff
@@ -67,10 +81,17 @@ export default {
       default: "ID",
       type: String,
     },
+    /** ID of session */
+    sessionId: {
+      default: "",
```
deepansh96: How about making the default as `null`?
And then checking in the code if `sessionId == null`?

src/pages/Home.vue
```diff
@@ -84,8 +105,13 @@ export default {
...
   },
   async created() {
-    /** Group name is sent to the API to retrieve all details */
-    this.groupData = await groupAPIService.getGroupData(this.group);
+    /** If sessionId exists in route, then retrieve session details. Otherwise, fallback is using group. */
+    if(this.sessionId === ""){
+      this.groupData = await groupAPIService.getGroupData(this.group); }
+    else{
+     this.sessionData = await sessionAPIService.getSessionData(this.sessionId);
+      (this.sessionData.enabled == "on") ? this.sessionEnabled = true : this.sessionEnabled = false;
```
deepansh96: Isn't sessionData.enabled a boolean? If it's not, shouldn't it be a boolean instead of a string?

deepansh96: You can re-write it like this
```suggestion
      this.sessionEnabled = (this.sessionData.enabled == "on") ? true : false;
```

deepansh96 (changes requested):

src/pages/Home.vue
sudheshna-donthineni: so actually the way it is stored from google sheets is "on" or "off". What I can do is while writing to firestore, I can store it as true or false.

deepansh96: yes that would be the correct way to do it, it removes dependence on strings.
if someone types "no " with a space, then we'll run into problems.

While writing to firestore, you can trim the string, make it lowercase and check if it's "yes" or "no" and save the value as boolean

src/components/SessionEntry.vue
deepansh96: Any update on this?

PR #52, opened 2022-04-07
Added session support

### LINKED ISSUE #53: Adding sessions support to auth layer

sudheshna-donthineni:
This code checks the timings of the session.


src/components/SessionEntry.vue
```diff
@@ -1,19 +1,111 @@
...
 </template>
 
 <script>
+import UserMessage from "./UserMessage.vue";
+import Entry from "./Entry.vue";
+import groupAPIService from "../services/API/groupData";
 
 export default {
   name: "SessionEntry",
+  components:{UserMessage, Entry},
```
deepansh96, line 26: linting!

deepansh96: If volar is not working to do linting, set up eslint for that

src/components/SessionEntry.vue
```diff
@@ -1,19 +1,111 @@
...
+
+       /** Retrieves destination ID */
+       getRedirectId(){
+         return this.sessionData.redirectPlatformParams.link
+       },
+
+       /** Retrieves group name */
+       getGroup(){
+         return this.sessionData.group
+       },
```
deepansh96: Either you should change the name of these. 
Because you're putting these values in computed properties, the names should not look like method names. 
Method is doing something, like "getting" something.
For example, for the group name, `getGroup` is a method. If you want to use it as a computed property, the name can simply be `groupName()`. According to the current code, how would you differentiate something like `getPurpose()` being in computed and `getSessionDateTimeInMilliseconds()` being in methods?
Both functions are doing the job of getting something.

Computed variables are not supposed to be used as method calls. They're supposed to be used as reactive variables. Let me know if it's not clear

Things like `isStartDateTimeValid()` is correct to be put in the computed properties.

deepansh96 (changes requested):

src/components/SessionEntry.vue
sudheshna-donthineni: My thinking was this, computed variables are cached, so I use them for operations that are least complex (to avoid bloating). Technically, `getSessionDateTimeInMilliseconds` could also be computed if it wasn't taking in a parameter. 

But yeah, I agree about the naming. Will change that.

sudheshna-donthineni: My thinking was this, computed variables are cached, so I use them for operations that are least complex (to avoid bloating). Technically, `getSessionDateTimeInMilliseconds` could also be computed if it wasn't taking in a parameter. 

But yeah, I agree about the naming. Will change that.

deepansh96: yes you're totally right about the segregation of logic among computed/methods
but just try to keep the naming consistent
computed properties' names should feel like variable names and not method names

PR #57, opened 2022-06-10
Checks for end date of the session

sudheshna-donthineni:
**Apologies for the unnecessary formatting changes. The eslint config is different across repos. Finding a solution to standarize across.**

This code adds two checks:
- Checks if `endDate` exists in the session data. If yes, check if the current date is less than or equal to the end date and combines the current date with the session end time to create the current day's end guard.  
- Added a check if `repeatSchedule` exists in the session data. If not, it is a non-repeating schedule and ignores the check.

### DIFF
--- a/src/components/SessionEntry.vue
+++ b/src/components/SessionEntry.vue
@@ -71,7 +71,7 @@ export default {
       return year + '-' + month + '-' + date;
     },
 
-    /** Checks if current date is greater than start date and combines session start date and start time */
+    /** Checks if current date is greater than start date and combines current date and start time */
     sessionStartDateTime() {
       return (
         this.sessionData.startDate <= this.currentDate &&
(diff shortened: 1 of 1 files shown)


src/components/SessionEntry.vue
```diff
@@ -31,119 +31,148 @@
...
-      return (new Date(this.currentDate + " " + this.sessionData.endTime))
+    /** Checks if current date is less than end date and combines current date and end time */
+    sessionEndDateTime() {
+      if ("endDate" in this.sessionData) {
+        return (
+          this.sessionData.endDate >= this.currentDate &&
+          new Date(this.currentDate + " " + this.sessionData.endTime)
+        );
+      }
+      return new Date(this.currentDate + " " + this.sessionData.endTime);
```
deepansh96: I didn't understand the purpose of putting an if check and the `this.sessionData.endDate >= this.currentDate &&` check because in both cases we are returning `return new Date(this.currentDate + " " + this.sessionData.endTime);`

sudheshna-donthineni: So the latter return is for sessions that don't have an endDate. In that case, we just build the datetime string and compare. The former is for when endDate is given. So it would return false if the current date has crossed the end date and will throw a no class message.

deepansh96 (approved):

pritamps (approved):

PR #58, opened 2022-06-20
Added navigation guard to the router

### LINKED ISSUE #59: Add navigation guard to the router
Add navigation guard to the router so that invalid query parameters added to the URL are detected.

sudheshna-donthineni:
- Added navigation guard to the router, which will route to a 404 page if the query parameters in the URL are invalid.

For now, following are the allowed query parameters: 
 ```
[
  "sessionId",
  "purpose",
  "subPurpose",
  "redirectTo",
  "redirectID",
  "group",
];
```

### CODE REVIEW

**Comment by pritamps** [comment]
Why redirectID and sessionId. Pick one :)

File: src/pages/Error.vue [hunk 1]
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div class="flex flex-col space-y-8 items-center">
+    <!-- 404 -->
+    <div v-if="isError404" class="w-full mt-16 flex flex-col space-y-4 items-center">
+      <inline-svg class="w-12 h-12 text-yellow-600 fill-current" :src="warningSvg" />
+
+      <p class="text-2xl align-middle">Page Not Found</p>
+      <p class="text-lg text-gray-500 text-center w-10/12 sm:w-1/2">
+        We are unable to find what you are looking for
```
pritamps: This page comes when required params are not passed, no? Maybe have a more descriptive error?

src/router/index.js
```diff
@@ -1,6 +1,16 @@
...
 import Home from "@/pages/Home.vue";
 import Sentry from "@/pages/Sentry.vue";
+import Error from "@/pages/Error.vue";
+
+const allowedQueryParams = [
+  "sessionId",
+  "purpose",
+  "subPurpose",
+  "redirectTo",
+  "redirectID",
```
pritamps, line 12: Fix the inconsistency between `sessionId` and `redirectID` . `Id` is better, probably?

pritamps (changes requested):

suryabulusu (approved):

src/router/index.js
pritamps: This is still `redirectID`? Shouldn't it be `redirectId`?

pritamps (changes requested):
Still `redirectID` in one place instead of `redirectId`?

src/router/index.js
sudheshna-donthineni: It's both, right? These are the list of allowed query parameters, so URLs, for now, can contain both `redirectID` or `redirectId`.

pritamps: If it's both, where is `redirectId`?

sudheshna-donthineni: There are two more params underneath our conversation 😛

pritamps: I see only `group` here below the conversation, not `redirectId` :-/

sudheshna-donthineni: Screenshot 2022-07-22 at 11 48 28 AM

PR #60, opened 2022-06-23
Time guards calculated on server

sudheshna-donthineni:
Before, we calculate the time guards on the client-side. This could result in falsey results as the user's device could be set in a different timezone. Also, different browsers represent Date object differently.

Now, the time guards are calculated on the server. The server only returns a boolean. This results in a uniform output.

Backend PR: https://github.com/avantifellows/auth-layer-backend/pull/8


src/components/SessionEntry.vue
```diff
@@ -1,178 +0,0 @@
-<template>
```
deepansh96, line 1: why was this deleted?

src/pages/Home.vue
```diff
@@ -112,25 +106,56 @@ export default {
...
-      this.groupData = await groupAPIService.getGroupData(this.group);
-    } else {
+    if (this.sessionId != null) {
       this.sessionData = await sessionAPIService.getSessionData(this.sessionId);
-      this.sessionEnabled = (this.sessionData.enabled == 1);
+      this.sessionEnabled = this.sessionData.sessionActive;
     }
-    this.isLoading = false
+    this.groupData = await groupAPIService.getGroupData(this.group);
+    this.isLoading = false;
```
deepansh96: in the earlier code version, one of two things was happening
Either we were getting group data or we were getting session data. But now both things are happening.
I didn't understand, why is that?

deepansh96 (changes requested):

src/components/SessionEntry.vue
sudheshna-donthineni: It's not needed anymore. All the session time computations are happening on the server side.

src/pages/Home.vue
sudheshna-donthineni: In the earlier code, both were happening, it was just a difference in which component. 
So the flow is this: If no `sessionId` is mentioned, then get `group` from URL and fetch group details. If `sessionId` is in the URL, then fetch the session details first, compute if the session is active, then fetch group details to show the UI. 

If `sessionId` is in the URL, the group details were fetched in the SessionEntry component. This time, it's fetched in the Home component itself. Group API is called regardless.

deepansh96 (approved):

PR #70, opened 2022-10-04
Send user's IP address

### LINKED ISSUE #71: Send user's IP address
Send every user's IP address to BQ.

sudheshna-donthineni:
- The IP address is sent from the backend because you can extract the address only through a request call. 
- `userNetworkData` is the object that contains all network data about the user. The next item will be the device details.

Test the below session and after redirection, check the BQ table. 
https://staging-auth.avantifellows.org/?sessionId=HaryanaStudents_B01_44672_vfr-mndk-ado
https://console.cloud.google.com/bigquery?authuser=0&project=avantifellows&ws=!1m10!1m4!4m3!1savantifellows!2sauth_logs!3sstaging-attendance-logs!1m4!1m3!1savantifellows!2sbquxjob_3dbcb634_183c099d63f!3sasia-south1

### DIFF
--- a/src/components/Entry.vue
+++ b/src/components/Entry.vue
@@ -126,6 +126,10 @@ export default {
       type: String,
       default: "",
     },
+    userData: {
+      type: Object,
+      default: {},
+    },
   },
   data() {
     return {
@@ -459,7 +463,8 @@ export default {
             this.authType,
             this.group,
             this.userType,
-            this.sessionId
+            this.sessionId,
+            this.userData
           );
         }
       }
--- a/src/components/Entry.vue
+++ b/src/components/Entry.vue
@@ -126,9 +126,9 @@ export default {
       type: String,
       default: "",
     },
-    userNetworkData: {
-      type: Object,
-      default: {},
+    userIpAddress: {
+      type: String,
+      default: "",
     },
   },
   data() {
@@ -464,7 +464,7 @@ export default {
             this.group,
             this.userType,
             this.sessionId,
-            this.userNetworkData
+            this.userIpAddress
           );
         }
       }
--- a/src/components/Entry.vue
+++ b/src/components/Entry.vue
@@ -126,7 +126,7 @@ export default {
       type: String,
       default: "",
     },
-    userData: {
+    userNetworkData: {
       type: Object,
       default: {},
     },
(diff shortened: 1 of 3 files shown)


src/components/Entry.vue
```diff
@@ -459,7 +463,8 @@ export default {
             this.authType,
             this.group,
             this.userType,
-            this.sessionId
+            this.sessionId,
+            this.userNetworkData
```
pritamps: This is only the IP address, no? Just call it IP address?

src/pages/Home.vue
```diff
@@ -148,6 +150,11 @@ export default {
         ? this.sessionData.purposeParams
         : this.purposeParams;
     },
+
+    /** Returns object containing network data about the user */
+    getUserNetworkData() {
```
pritamps: Same comment. Since only IP, method name should reflect that

pritamps (changes requested):

src/pages/Home.vue
deepansh96: Agreed
If you're planning to store more information, then that will be network data AND other user-agent data
So then it makes sense to store it in an object but then it should not be called `networkData`

If you're storing JUST the network data, where I think IP address is the only thing we need, then the name should also reflect that

sudheshna-donthineni: So the reason I named it this was, for now it's only IP address, but in the future we could want to capture extra data, so created an object to which we can add furthur fields.

User agent data is also network data no?

deepansh96: Uhh not exactly
But then again it's subjective, depends on person to person
Maybe we can call it user meta data?

or even better, you can have userIP address and user device data
IP address will be just a string
device data will be an object cause it will contain many things

pritamps: No no. Just call it IP address since that's what we care about. The method name should reflect what we're trying to get out. If we need more things later, we can either add more methods, or add a `getNetworkData` method and submethods which will return individual values. Right now, the issue is that someone reading the method name won't know that what we care about is the IP. Makes sense?

src/components/Entry.vue
pritamps: Same here. We care only about IP, so should pass only IP

pritamps (approved):
Looks good to me!

PR #73, opened 2022-11-07
Multiple authentication methods

sudheshna-donthineni:
- One change in how the group data is being stored - we will be storing a string called `authType` which will contain the different authentication types the group will use. 

Screenshot 2022-11-07 at 2 53 10 PM

- Sample link : https://lnk.avantifellows.org/Du3N


src/pages/Home.vue
```diff
@@ -108,12 +105,12 @@ export default {
   computed: {
     /** Whether authentication method chosen is an ID entry */
     isAuthTypeID() {
```
pritamps, line 114: I feel like this is a good time to combine all this into a single `getAuthType` method and not have to rely on separate methods for `id`, `OTP`, etc.

Maybe an enum with supported auth types and their related info?

src/pages/Home.vue
```diff
@@ -154,6 +151,16 @@ export default {
...
     },
+
+    /** Apart from ID, are any extra inputs being validated. */
+    isExtraInputValidationsRequired() {
+      return this.getAuthType.split(",").length > 1;
+    },
+
+    /** Returns the auth methods used by each group */
+    getAuthType() {
+      return this.groupData && this.groupData.authType ? this.groupData.authType : "ID";
```
pritamps, line 184: Not sure what's happening -- what is the logic here?

src/services/API/user.js
```diff
@@ -0,0 +1,59 @@
+import { client } from "@/services/API/rootClient.js";
```
pritamps, line 1: This is a new file? Why is it needed?

src/services/validation.js
```diff
@@ -66,28 +90,51 @@ async function checkUserIdInFirestore(
...
+          userID,
+          validateCount,
+          dataSource["name"],
+          dataSource["column"],
+          isExtraInputValidationRequired
+        )
       );
+    } else {
+      if (authType.includes("ID")) {
+        return checkUserIdInFirestore(
```
pritamps, line 117: How come even if the `datasource` is not `Firestore`, we're still checking if the `id` is in `Firestore`?

src/services/API/user.js
sudheshna-donthineni: It's not a new file by logic. I just renamed it. It was called `checkUser.js`. Now renamed it to `user.js` since all the user related APIs are here.

src/pages/Home.vue
sudheshna-donthineni: The comment you mentioned above; combined it all into one `getAuthType`. That's what is happening. In case, the group does not have any information regarding what auth type to use, it defaults to ID.

src/services/validation.js
sudheshna-donthineni: So, if you look at the code, there is no checking for another data source. I'll make the change.

src/pages/Home.vue
sudheshna-donthineni: That's what is done below. But the reason I have two different methods for `ID` and `OTP` auth type is because they have different components to render.

pritamps: Should be combined, yo. Can you think of a way to combine them? Components can be part of a return data structure. 

It doesn't have to be in this PR, but make a note and after this ships, we can prioritize it, because this will get messier with time

src/pages/Home.vue
pritamps: I'm not sure I get the `if` condition, though. If `groupData` and `groupData.authType` exist, it returns `ID`, because that's what it looks like is happening

src/services/validation.js
pritamps: yes, just code prettiness and understandability change needed here

src/pages/Home.vue
sudheshna-donthineni: No, if `groupData` and `groupData.authType` exists, then `groupData.authType` is returned, otherwise `ID` is defaulted.

src/services/validation.js
```diff
@@ -66,28 +90,51 @@ async function checkUserIdInFirestore(
...
-      );
-    }
-    if (authType == "OTP") {
-      return checkPhoneNumberInFirestore(
-        userID,
-        dataSource["name"],
-        dataSource["column"]
+    if (isExtraInputValidationRequired) {
+      return (
+        checkBirthdateInFirestore(
```
suryabulusu: why not a single function to check birthdate+user_id combination? either the pair should exist together -- or we do not allow logging-in.

src/components/Entry.vue
```diff
@@ -143,13 +187,34 @@ export default {
...
+      ],
     };
   },
   created() {
     /** The user type is set as soon as component is created */
     this.userType = this.groupData.userType;
   },
   computed: {
+    isExtraInputValidationRequired() {
+      return this.authType.split(",").length > 1;
```
suryabulusu: whats this `.split(",").length > 1` part?

sudheshna-donthineni: This tells us if there is more than one input that needs to be validated.

src/pages/Home.vue
sudheshna-donthineni: Made an issue.

src/services/validation.js
sudheshna-donthineni: Actually, I misunderstood your comment. This else statement is inside `dataSource[type] == 'Firestore'`. So I think it makes sense, no?

src/components/Entry.vue
```diff
@@ -8,13 +8,20 @@
       />
     </div>
   </div>
+  <div class="flex w-full h-10 justify-evenly md:w-5/6 md:h-20 xl:w-3/4 mx-auto mt-20">
+    <template v-for="(image, index) in getGroupImages" :key="index">
+      <img :src="image" />
+    </template>
```
deepansh96, line 14: just curious, why `template` here?
I tried looking on the docs but could not find the reason

src/components/Entry.vue
```diff
@@ -53,9 +60,9 @@
     <span v-if="isInvalidInputMessageShown" class="mx-auto text-red-700 text-base mb-1">{{
       invalidInputMessage
     }}</span>
-    <span v-if="isInvalidLoginMessageShown" class="mx-auto text-red-700 text-base mb-1">{{
+    <!-- <span v-if="isInvalidLoginMessageShown" class="mx-auto text-red-700 text-base mb-1">{{
       invalidLoginMessage
-    }}</span>
+    }}</span> -->
```
deepansh96: is this needed? if not, can you remove?

src/components/Entry.vue
```diff
@@ -70,15 +77,75 @@
...
       </button>
     </div>
+    <div v-show="isExtraInputValidationRequired" class="flex flex-col pt-10 pb-6">
+      <div>
+        <p
+          class="w-1/2 text-xl lg:text-2xl xl:text-3xl mx-auto font-bold md:w-3/4 text-center"
+        >
+          Enter your birthdate
+        </p>
+      </div>
```
deepansh96, line 89: shouldn't this also be in the group data JSON on firestore?

src/components/Entry.vue
```diff
@@ -143,13 +215,27 @@ export default {
...
+    this.$store.dispatch("setGroupData", this.groupData);
   },
   computed: {
+    getGroupImages() {
+      return this.groupData.images;
+    },
+    /** Returns if more than one input needs to be validated */
+    isExtraInputValidationRequired() {
+      return this.authType.split(",").length > 1;
+    },
```
deepansh96: Along with checking the length, we should also check if the split entities are not empty strings
I know that we would never put such a value in the firestore or DB, but still those kind of cases should be handled gracefully here
So for an auth type like `DOB,` or `,` or `,ID`, this will be wrong because we don't have the complete information

Acha do we have stored anywhere the types of valid auth types?

src/components/Entry.vue
```diff
@@ -201,7 +287,8 @@ export default {
       return (
         !this.isAnyUserIDPresent ||
         this.invalidInputMessage != "" ||
-        this.isCurrentEntryIncomplete
+        this.isCurrentEntryIncomplete ||
+        this.isBirthDateEntryIncomplete
```
deepansh96, line 288: for later, maybe not now, but what would be good to have is a separate method which we can call, `areAllRequiredFieldsFilled` or something like that which would check if all required fields are filled up properly or not and then we can call that method (or computed var) here

src/components/Entry.vue
```diff
@@ -231,7 +327,10 @@ export default {
 
     /** Whether the current typed ID is valid */
     isInvalidLoginMessageShown() {
-      return !this.isCurrentUserValid && this.validateCount == 1;
+      return (
+        (!this.isCurrentUserValid && this.validateCount == 1) ||
+        (!this.isCurrentUserValid && this.validateCount == 0)
+      );
```
deepansh96: why is this being done?

src/pages/Home.vue
```diff
@@ -21,9 +21,10 @@
         :purposeParams="getPurposeParams"
         :groupData="groupData"
         :group="getGroup"
-        :authType="authType"
+        :authType="getAuthType"
         :sessionId="sessionId"
         :userIpAddress="getUserIpAddress"
+        :extraInputValidation="isExtraInputValidationsRequired"
```
deepansh96: can you also keep the prop's name to `isExtraInputValidationsRequired`?
because both variables are actually boolean, so it makes it easier for someone reading the code to understand that it is a boolean variable

src/pages/Home.vue
```diff
@@ -154,6 +151,16 @@ export default {
     getUserIpAddress() {
       return this.sessionData ? this.sessionData.userIp : "";
     },
+
+    /** Apart from ID, are any extra inputs being validated. */
+    isExtraInputValidationsRequired() {
+      return this.getAuthType.split(",").length > 1;
```
deepansh96: If this thing is being calculated at Home, and you're sending the value of this computed variable to the child `Entry`, then why aren't you using this value inside `Entry`?
Inside `Entry`, I saw that you're recalculating this information using `this.authType`

src/router/index.js
```diff
@@ -27,7 +27,6 @@ const routes = [
       // legacy URLs support redirectID, new URLs must use redirectId
       redirectId: route.query.redirectID || route.query.redirectId,
       group: route.query.group,
-      authType: route.query.authType,
```
deepansh96, line 30: why was this removed?
because the authType will be available in session/group data is that why?
If yes, then what about the older links which had `authType` as param? Will they be working fine?

src/services/API/user.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Creates a new student
+ * @param {Object} formData - contains data filled in the form by user
+ */
+export default {
+  studentSignup(formData) {
+    return new Promise((resolve) => {
+      client
+        .post(studentSignupEndpoint, JSON.stringify(formData))
```
deepansh96, line 16: where is the `studentSingupEndpoint` variable imported? I can see its not imported from the endpoints file

deepansh96 (changes requested):

src/router/index.js
sudheshna-donthineni: Yeah, and also, I had a fallback value in the code. 

Screenshot 2022-11-16 at 11 46 58 AM

src/services/API/user.js
sudheshna-donthineni: I always have trouble with multiple connected PRs. This happens in the `signup` PR.

src/pages/Home.vue
sudheshna-donthineni: Correct, made the change.

src/pages/Home.vue
sudheshna-donthineni: Done.

src/components/Entry.vue
sudheshna-donthineni: Like discussed in call, added a hack check in the `signup` PR.

src/components/Entry.vue
sudheshna-donthineni: So this is something I learnt recently and started using it, especially when it is loops and conditional loops (using `<template>` instead of `<div>`) There is no performance difference, but `<template>` does not show up on the DOM, unlike `<div>`. This will help if you have too many conditional renders to avoid many `<div>`s in the markup. Also helps with styling, does not add unnecessary margins.

src/components/Entry.vue
sudheshna-donthineni: So actually made a change in the `signup` PR.

src/components/Entry.vue
sudheshna-donthineni: Ideally, yes. But haven't figured out the correct format for the JSON object in cases where there are additional input parameters. Have to think about it.

src/components/Entry.vue
sudheshna-donthineni: Made the change in the `signup` PR. 

Not really, for now I just made a global variable.

suryabulusu (approved):

src/components/Entry.vue
deepansh96: got it
so maybe next big task for auth layer could be to
- learn from this sign up page stint and make some strings configurable which are hardcoded as of now
- and in general try to make auth layer such that adding a new auth profile should be automated, and not manual. like there should and must be a general way to add a new authentication type

deepansh96 (approved):

src/services/validation.js
```diff
@@ -45,49 +66,76 @@ async function checkUserIdInFirestore(
...
         dataSource["name"],
         dataSource["column"]
       );
+      // let isIdValid = await checkUserIdInFirestore(
+      //   userID,
+      //   validateCount,
+      //   dataSource["name"],
+      //   dataSource["column"],
+      //   isExtraInputValidationRequired
+      // );
```
deepansh96: what about this?

Issue #84, opened 2023-02-01
Include JWT authentication


sudheshna-donthineni:
- Currently, the way Portal redirects to other Avanti services is not very secure. API keys and user IDs are sent in the form of URL query parameters.
- To secure this process, a JWT authentication service has been built under the repo `portal-backend`. 
- Need to incorporate this auth system into Portal such that before a user is redirected, a token is generated with all the sensitive information and this token will be used by other Avanti services as well.

Akshanshkaushal:
@sudheshna-donthineni  @suryabulusu I would love to work upon this issue. Could you please assign this to me ?

PR #85, opened 2023-02-02
Incorporating jwt authentication in portal

Bahugunajii:
## Target
#84 

## Description
Here we are sending a request to the portal backend to generate an access token with the sensitive information included. The access token is then stored in the cookie for other Avanti services to verify.


src/services/API/endpoints.js
```diff
@@ -5,3 +5,5 @@ export const verifyOTPEndpoint = "/verifyotp";
 export const getSessionDataEndpoint = "/getSessionData";
 export const checkBirthdateEndpoint = "/checkBirthdate";
 export const studentSignupEndpoint = "/studentSignup";
+export const createAccessTokenEndpoint =
+  import.meta.env.VITE_APP_PORTAL_BACKEND + "/create-access-token";
```
sudheshna-donthineni: this is an endpoint so only the API endpoint must be mentioned. The base URL can be called from where you making the API call.

src/services/API/setToken.js
```diff
@@ -0,0 +1,17 @@
+import { createAccessTokenEndpoint } from "./endpoints";
+
+export function setToken(apiKey, unique_id) {
```
sudheshna-donthineni: If the arguments are fixed, this function can't be reusable. Every destination does not use `apiKey` and `unique_id`, no.

src/services/API/setToken.js
```diff
@@ -0,0 +1,17 @@
...
+  return fetch(createAccessTokenEndpoint, {
```
sudheshna-donthineni: Also, if you look at the other APIs, we use axios for API calling.

src/services/redirectToDestination.js
```diff
@@ -35,12 +36,24 @@ export function redirectToDestination(
     }
     case "quiz": {
       redirectURL = import.meta.env.VITE_APP_BASE_URL_QUIZ;
+      const apiKey = import.meta.env.VITE_APP_QUIZ_AF_API_KEY;
+      const unique_id = userID;
```
sudheshna-donthineni: Don't rename the variable. It's called `userId`, let it be the same.

src/services/API/setToken.js
Bahugunajii: There will be two cases for this:
1) The destination independent of `apiKey` and `unique_id` will not use this function
2) For the destinations having additional fields other than `apiKey` and `unique_id` we can have an argument named data from where apiKey, unique_id and similar other fields can be extracted. Currently, I don't see any destination with such case, but surely It can be modified when we would have such a condition

src/services/API/setToken.js
Bahugunajii: done

src/services/redirectToDestination.js
Bahugunajii: sure

sudheshna-donthineni:
Closing this for now. When we get to the auth API migration, will reopen.

PR #86, opened 2023-02-10
Phone number support for Delhi Sessions

sudheshna-donthineni:
Changes made:
- Added a new component for phone number entry. This is the intent moving forward. Everytime a new method is added, the parent component will be the `Entry.vue` which will contain any common methods and each auth method will have a component of its own with its own logic. The birthdate component will be moved next.
- `ref` is being used for accessing the child's methods. 
- Until now mutliple IDs were not allowed only if the destination was plio or quiz. Now on, if we are using multiple methods as once, the add button will be disabled as well.
- Also, don't why some formatting happened. :|

For testing:

https://staging-auth.avantifellows.org/?sessionId=NEET__62c29f2573d3bb41aa88554d

### DIFF
--- a/src/components/Entry.vue
+++ b/src/components/Entry.vue
@@ -8,7 +8,9 @@
       />
     </div>
   </div>
-  <div class="flex w-11/12 h-10 justify-evenly md:w-5/6 md:h-20 xl:w-3/4 mx-auto mt-20">
+  <div
+    class="flex w-11/12 h-10 justify-evenly md:w-5/6 md:h-20 xl:w-3/4 mx-auto mt-20"
+  >
     <template v-for="(image, index) in getGroupImages" :key="index">
       <img :src="image" />
     </template>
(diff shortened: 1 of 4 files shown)


src/components/Entry.vue
```diff
@@ -502,17 +536,18 @@ export default {
       );
       if (this.isExtraInputValidationRequired) {
         this.isCurrentUserValid = userValidationResponse;
+        this.validateCount = 2;
```
sudheshna-donthineni, line 539: This bypasses the second validation call. Will think of a more sophisticated method.

sudheshna-donthineni:
For SQS, whenever a new data variable needs to logged into BQ, it is being appended to the message and I'm manually adding a new column to the table. For SQS, I'm thinking all the data variables can become an object, so that new variables will become easier to add. For adding a column into the BQ table, I'm not sure how to automate it except to make the data type as JSON. But I guess it's not easy to query such columns.

src/components/Entry.vue
pritamps: Yes this is not good

src/components/Entry.vue
```diff
@@ -225,6 +248,14 @@ export default {
     };
   },
   computed: {
+    /** Returns if the authentication type includes date of birth */
+    isInputDateOfBirth() {
```
pritamps, line 252: This also seems a bit off to me.. a variable and a method for all the things. There should be a more direct relation with the group parameters... But we can look at it together later

src/components/Entry.vue
```diff
@@ -524,7 +560,7 @@ export default {
       await this.authenticateID(latestUserID);
       if (!this.isCurrentUserValid && this.validateCount == 0) {
         this.invalidLoginMessage =
-          "Student ID/Birthdate entered is incorrect. Please try again!<br/> Please register incase you are a new user.";
+          "Student ID entered is incorrect. Please try again!<br/> Please register incase you are a new user.";
```
pritamps, line 563: Registration form is there for these guys?

What about the contact number to call?

src/components/PhoneNumberEntry.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+  <div class="flex flex-col justify-center">
+    <p
+      class="w-1/2 text-xl lg:text-2xl xl:text-3xl mx-auto font-bold md:w-3/4 text-center"
+    >
```
pritamps, line 5: This is a new component?

pritamps: Got it

pritamps (approved):
Lots of things to think about. But we'll talk together later

PR #89, opened 2023-02-23
Connect to new backend

sudheshna-donthineni:


pritamps:
What is the strategy here? Once this is deployed, everything is connected to the new DB, right? Are we going to switch overnight -- how will we make sure all current sessions are in both places so no students lose any classes?

sudheshna-donthineni:
Closing this because we are migrating one API at a time now.

PR #93, opened 2023-05-03
Migration from old user API to new API

sudheshna-donthineni:
First step in migrating our old backend to the new.
The only main change is that we won't have two different APIs for ID and anything else. 
Instead, we have only one API which will validate whatever student params we send it.

Test Cases:
[DelhiStudents](https://staging-auth.avantifellows.org/?sessionId=DelhiStudents_11B42_44827_ubm-pczd-ntn)
[EnableStudents](https://staging-auth.avantifellows.org/?sessionId=EnableStudents_B02_44801)
[HimachalStudents]()

### DIFF
--- a/.github/workflows/deploy_to_s3_prod.yml
+++ b/.github/workflows/deploy_to_s3_prod.yml
@@ -15,50 +15,50 @@ jobs:
       url: "https://auth.avantifellows.org"
 
     steps:
-    - name: Checkout
-      uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
 
-    - name: Cache node modules
-      uses: actions/cache@v1
-      with:
-        path: node_modules
-        key: ${{ runner.OS }}-build-${{ hashFiles('**/package-lock.json') }}
-        restore-keys: |
-          ${{ runner.OS }}-build-
-          ${{ runner.OS }}-
+      - name: Cache node modules
+        uses: actions/cache@v1
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-build-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.OS }}-build-
+            ${{ runner.OS }}-
 
-    - name: Install
-      run: npm install
+      - name: Install
+        run: npm install
 
-    # build the vuepress docs
-    - name: Build
-      env:
-        NODE_ENV: production
-        VITE_APP_PLIO_AF_API_KEY: ${{ secrets.VITE_APP_PLIO_AF_API_KEY }}
-        VITE_APP_BASE_URL_PLIO: 'https://app.plio.in/#/af/play/'
-        VITE_APP_QUIZ_AF_API_KEY: ${{ secrets.VITE_APP_QUIZ_AF_API_KEY }}
-        VITE_APP_BASE_URL_QUIZ: 'https://quiz.avantifellows.org/quiz/'
-        VITE_APP_AWS_ACCESS_KEY_ID: ${{ secrets.VITE_APP_AWS_SQS_ACCESS_KEY }}
-        VITE_APP_AWS_SECRET_KEY: ${{ secrets.VITE_APP_AWS_SQS_SECRET_ACCESS_KEY }}
-        VITE_APP_BASE_URL_MEET: 'https://meet.google.com/'
-        VITE_APP_OTP_SERVICE: ${{ secrets.VITE_APP_OTP_SERVICE }}
-        VITE_APP_AWS_REGION: ${{ secrets.VITE_APP_AWS_REGION }}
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy_to_s3_prod.yml
```diff
@@ -15,49 +15,50 @@ jobs:
...
-      uses: aws-actions/configure-aws-credentials@v1
-      with:
-        aws-access-key-id: ${{ secrets.AWS_S3_ACCESS_KEY_ID }}
-        aws-secret-access-key: ${{ secrets.AWS_S3_SECRET_ACCESS_KEY }}
-        aws-region: ap-south-1
-
-    # sync with the given S3 bucket
-    - name: Deploy static site to S3 bucket
-      run: aws s3 sync ./dist s3://af-auth-layer --delete --acl public-read
+      - name: Checkout
```
pritamps, line 18: Why did you change the indentation? Doesn't seem to be changed for other sections

.github/workflows/deploy_to_s3_staging.yml
```diff
@@ -16,49 +16,50 @@ jobs:
       url: "https://staging-auth.avantifellows.org"
 
     steps:
-    - name: Checkout
-      uses: actions/checkout@v2
+      - name: Checkout
```
pritamps, line 19: Same question here

src/components/Entry.vue
```diff
@@ -525,14 +525,13 @@ export default {
...
-
       let userValidationResponse = await validateID(
         userID,
-        this.groupData.dataSource,
         this.authType,
         this.validateCount,
         this.dateOfBirth,
-        this.isExtraInputValidationRequired
+        this.isExtraInputValidationRequired,
+        this.$refs.phoneNumberEntry.phoneNumber
```
pritamps, line 534: Why is `phoneNumberEntry` coming from `$refs` unlike the others?

src/services/API/user.js
```diff
@@ -1,9 +1,8 @@
-import { client } from "@/services/API/rootClient.js";
+import { client, newClient } from "@/services/API/rootClient.js";
```
pritamps: `newClient` is a very bad name. Call it `AFDBClient` maybe? Or something descriptive

src/services/validation.js
```diff
@@ -88,47 +42,56 @@ async function checkUserIdInFirestore(
...
-      if (authType.includes("OTP")) {
-        return checkPhoneNumberInFirestore(
-          userID,
-          dataSource["name"],
-          dataSource["column"]
-        );
+      if (userValidationResponse) {
+        if (authType.includes("DOB")) {
+          let dob =
+            (birthdate.day < 10 ? "0" + birthdate.day : birthdate.day) +
```
pritamps: Are we 100% sure days<10 will be stored **without** the leading 0?

src/services/validation.js
```diff
@@ -88,47 +42,56 @@ async function checkUserIdInFirestore(
...
+            "-" +
+            birthdate.year;
+          var isBirthdateValid = await userAPI.verifyStudent({
+            student_id: userID,
+            birth_date: dob,
+          });
+          return userValidationResponse && isBirthdateValid;
+        }
+        if (authType.includes("PH")) {
+          console.log(authType);
```
pritamps: Remove `console.log`

src/services/validation.js
```diff
@@ -88,47 +42,56 @@ async function checkUserIdInFirestore(
...
-          userID,
-          dataSource["name"],
-          dataSource["column"]
-        );
+      if (userValidationResponse) {
+        if (authType.includes("DOB")) {
+          let dob =
+            (birthdate.day < 10 ? "0" + birthdate.day : birthdate.day) +
+            "-" +
+            (birthdate.month < 10 ? "0" + birthdate.month : birthdate.month) +
```
pritamps, line 70: Same question for year.

pritamps (changes requested):

.github/workflows/deploy_to_s3_prod.yml
sudheshna-donthineni: This seems to have happened with `pre-commit`.

src/components/Entry.vue
sudheshna-donthineni: Because `phoneNumberEntry` is a separate component. The ID and DOB are in the same `Entry` component. Moving forward I am creating a separate component for each kind of entry type, so that they can be mixed and matched for different forms.

src/services/validation.js
sudheshna-donthineni: They are stored with the leading 0. The entry does not have a leading 0, so to compare we append 0 before sending to the backend.

pritamps (approved):
Looks good to me!

suryabulusu (commented):
looks good!

PR #94, opened 2023-05-16
Sign-in flow

sudheshna-donthineni:
This is the new sign-in flow. Much more simple and streamlined. 
We don't have `validateCount` anymore because we are not letting students in if they enter incorrect ID.

https://staging-auth.avantifellows.org/?sessionId=portal_v2

ID - 1111111111, Phone - 9999999999, DOB - 2001-01-01

For more context: https://www.notion.so/avantifellows/ADR-Portal-2-0-8cb08f877fcb42cc920cd010cee88d62?pvs=4


src/components/Datepicker.vue
```diff
@@ -0,0 +1,53 @@
...
+      default: "",
+    },
+  },
+  data() {
+    return {
+      date: "",
+    };
+  },
+  computed: {
+    isDateEntryCompleteAndValid() {
```
suryabulusu: the function name is a bit off, no? If date is required, we check if its not empty and return true.

If date is not required, it is fine if date is empty. We return true regardless. But its not "complete" or "valid" by being empty.

src/components/PhoneNumberEntry.vue
```diff
@@ -53,22 +78,19 @@ export default {
...
-    isPhoneNumberNotComplete() {
-      return this.phoneNumber.length < 10;
-    },
+
     /** Updates phone number based on user entry */
-    updatePhoneNumber(event) {
+    updatePhoneNumberEntry(event) {
       if (event.target.value.length == 0) {
         this.invalidPhoneNumberMessage = "";
       } else if (event.target.value.length > 10) {
```
suryabulusu: maxPhoneNumberLength = 10 we can set a variable

src/components/TextEntry.vue
```diff
@@ -0,0 +1,90 @@
...
+      return [
+        {
+          "border-red-600 focus:border-red-600": this.invalidTextEntryMessage,
+        },
+      ];
+    },
+    /** Updates text based on user entry */
+    updateTextEntry(event) {
+      if (event.target.value.length == 0) {
+        this.invalidTextEntryMessage = "";
```
suryabulusu, line 88: we can have a maxTextLength here na just like number and phonenumber

src/pages/Signin.vue
```diff
@@ -0,0 +1,259 @@
...
+import { redirectToDestination } from "../services/redirectToDestination";
+import { sendSQSMessage } from "@/services/API/sqs";
+
+const assets = useAssets();
+export default {
+  name: "SignIn",
+  components: {
+    NumberEntry,
+    Datepicker,
+    PhoneNumberEntry,
```
suryabulusu: no textentry?

src/components/TextEntry.vue
sudheshna-donthineni: How would you decide what the upper limit would be? Text does not limits. In fact, it can a min limit.

src/components/PhoneNumberEntry.vue
sudheshna-donthineni: It's a phone number. That's the component. So it's always going to be 10.

src/pages/Signin.vue
sudheshna-donthineni: `SignIn` will not have `TextEntry`, at least for now. Our only validators are those three. Until further is decided.

suryabulusu (approved):
looks good to me!

src/pages/Home.vue
```diff
@@ -113,67 +115,31 @@ export default {
...
-    getPurposeParams() {
-      return this.purposeParams == ""
-        ? this.sessionData.purposeParams
-        : this.purposeParams;
+    isPurposeRegistration() {
+      return this.sessionData ? this.getPurpose == "registration" : false;
     },
 
-    /** Returns IP address of user */
+    /** TEMP - Returns IP address of user */
```
suryabulusu, line 127: why temporary

PR #95, opened 2023-05-22
Sign up flow

sudheshna-donthineni:
This is the sign up flow for Portal V2. 
The form data is dummy right now. The form templates are still being finalized so using this data for now.
https://staging-auth.avantifellows.org/?sessionId=signup_v2


src/pages/Signup.vue
```diff
@@ -0,0 +1,200 @@
...
+      :placeholder="formField.placeholder"
+      :options="formField.options"
+      :multiple="formField.multiple"
+      :maxLengthOfEntry="formField.maxLengthOfEntry"
+      @update="updateUserData"
+    />
+    <!-- submit button -->
+    <button
+      class="bg-primary hover:bg-primary-hover text-white font-bold shadow-xl uppercase text-lg mx-auto p-4 rounded disabled:opacity-50 btn"
+      :disabled="signUpDisabled"
```
sudheshna-donthineni: This being a computed variable is not being reactive for me.

src/components/formData.json
```diff
@@ -0,0 +1,91 @@
...
+                        "dependantField": "region",
+                        "dependantFieldMapping":{
+                                "Bhopal": ["Chhattisgarh", "Madhya Pradesh", "Odisha"],
+                                "Lucknow": ["Uttarakhand", "Uttar Pradesh"],
+                                "Chandigarh": [
+                                  "Chandigarh",
+                                  "Himachal Pradesh",
+                                  "Jammu & Kashmir",
+                                  "Ladakh",
+                                  "Punjab"
```
suryabulusu, line 56: do we need to put this in every form that uses state info in JNV? what about state -> jnv district mapping? can't we use some fixed json for these mappings

src/services/API/user.js
```diff
@@ -11,16 +11,24 @@ import { checkForUserEndpoint } from "./endpoints";
...
+            form_data: formData,
+            id_generation: idGeneration,
+            user_type: "student",
+          })
+        )
         .then((response) => {
           resolve(response.data.toString());
         })
         .catch((error) => {
+          console.log(error);
```
suryabulusu, line 29: do we need this in prod

suryabulusu (commented):
looks good to me -- left a comment

src/services/API/user.js
sudheshna-donthineni: nope, will remove it.

PR #96, opened 2023-06-05
Added reporting engine as destination

sudheshna-donthineni:
Have an auth layer page before accessing reports so that students can enter their IDs. 

Example staging link: https://lnk.avantifellows.org/U9sh
Valid user id: 23031068442

If wrong ID is entered, show an error page [todo in reporting repo]


pritamps (approved):
Small change, and then looks good!

PR #97, opened 2023-06-06
candidate logic

sudheshna-donthineni:
Hack logic  because this needs to be blocking. So no matter how many times they enter incorrect ID, the validation count is set to 0. So validation is in a loop until the correct ID is entered.


suryabulusu (approved):
do remove this hacky logic and have something general later on

PR #127, opened 2023-12-14
Gurukul Support & supporting non-session Portal links & token auth

sudheshna-donthineni:
- Added query parameters
- Changed variable names
- Lifted certain computed methods from child to parent (`NewSignIn` to `Home`)
- Also, stores access token and refresh token in cookies


src/pages/Home.vue
```diff
@@ -175,32 +268,98 @@ export default {
...
+        (this.auth_type && this.auth_type.split(",")) ||
+        (this.groupData && this.groupData.input_schema.auth_type.split(","))
+      );
+    },
+
+    /** Returns if sign up flow should be enabled */
+    isSignupEnabled() {
+      return (
+        (this.sessionData && this.sessionData.enable_signup == "True") ||
+        this.enable_signup == "True"
```
suryabulusu, line 277: when will this.enable_signup = True happen? its not being set anywhere?

src/pages/NewSignup.vue
```diff
@@ -300,16 +305,16 @@ export default {
...
       ) {
         sendSQSMessage(
           "attendance-sign-up",
           this.$store.state.sessionData.purpose["sub-type"],
-          this.$store.state.sessionData.platform,
-          this.$store.state.sessionData.platform_id,
+          this.platform,
+          this.platform_id,
           this.userData["user_id"],
           "",
```
suryabulusu, line 295: what is this for? can u add besides it in comment

src/pages/Home.vue
sudheshna-donthineni: Its a URL parameter.

src/pages/NewSignup.vue
sudheshna-donthineni: Done

PR #138, opened 2024-04-10
Flatten sqs structure

sudheshna-donthineni:


src/services/API/sqs.js
```diff
@@ -43,19 +43,14 @@ export async function sendSQSMessage(
...
-        subType: purposeParams,
-        params: {
-          platform: redirectTo,
-          id: redirectId,
-        },
-      },
+      type: type,
+      sub_type: sub_type,
+      platform: platform,
+      platformId: platformId,
```
suryabulusu: shouldnt this be platform_id

src/services/API/sqs.js
```diff
@@ -43,19 +43,14 @@ export async function sendSQSMessage(
...
       authType: authType,
```
suryabulusu: auth_type

suryabulusu: and similar for rest

suryabulusu (commented):
left comments

suryabulusu (approved):

PR #142, opened 2024-05-25
Writing auth logs to PG

sudheshna-donthineni:


src/pages/NewSignin.vue
```diff
@@ -362,6 +363,12 @@ export default {
 
         if (this.enable_popup) {
           this.$router.push(`/form/${this.userInformation["student_id"]}`);
+          UserAPI.postUserSessionActivity(
+            this.userInformation["student_id"],
+            "sign-in",
```
pritamps, line 368: Do we have a constants file? If not we should create it and move all these strings there so they are all in one place

pritamps (changes requested):
Just needs a string constants file, otherwise looks good!

src/pages/NewSignin.vue
sudheshna-donthineni: Do we need a separate file? It's just one string per component/flow.

PR #151, opened 2024-07-25
Refactors Validation code

suryabulusu:


pritamps (approved):

src/services/newValidation.js
```diff
@@ -8,43 +8,53 @@ export async function validateUser(
...
-        auth_group_id: authGroupId,
-      });
-    }
-    if (userType == "teacher") {
-      user["isUserIdValid"] = await userAPI.verifyTeacher({
-        teacher_id: userInformation["teacher_id"],
-      });
-    }
-  }
+  let studentVerificationParams = {
```
pritamps: Will `userInformation` always contain `student_id`?

pritamps (changes requested):

src/services/newValidation.js
suryabulusu: not always. as of now returning `undefined` for non student logins . .silently failing. will change code

PR #152, opened 2024-09-05
Move AFTesting to V2

suryabulusu:
- Added AFTesting product
- Added AFTesting batches
- Added AFTesting authgroup 

Check here: [https://staging-auth.avantifellows.org?sessionId=AFTesting_66d9f20c494fc4cf6d7fe9fe](https://staging-auth.avantifellows.org/?sessionId=AFTesting_66d9f20c494fc4cf6d7fe9fe)

### DIFF
--- a/src/pages/NewSignin.vue
+++ b/src/pages/NewSignin.vue
@@ -383,6 +383,13 @@ export default {
      * @returns {Promise<void>} A Promise that resolves once the authentication process is complete.
      */
     async authenticate() {
+      var TESTING_MODE = false;
+      if (this.$store.state.authGroupData.name == "AFTesting") {
+        TESTING_MODE = true;
+        // authenticate all userIds as valid
+        // do not send logs to afdb
+      }
+
       let isUserValid = await validateUser(
         this.auth_type,
         this.userInformation,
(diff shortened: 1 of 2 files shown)


src/pages/NewSignin.vue
```diff
@@ -459,8 +463,8 @@ export default {
           );
           this.$router.push(`/form/${this.userInformation["student_id"]}`);
         } else {
-          if (this.$store.state.sessionData.session_id != null) {
-              // do not send logs for reports, gurukul
+          if (this.$store.state.sessionData.session_id != null && this.$store.state.authGroupData.name != "AFTesting") {
```
pritamps: Maybe have a global variable called "TESTING_ONLY" at the top of the file which is true only for `AFTesting` now, but can be expanded later. Will make it clearer what is going on to someone else who will read the code, and will also allow us to add different sort of testing etc with similar flow

Also add a comment when you define the global variable about the implications: (no ID checking, no data collection)

pritamps (changes requested):
Small suggestion!

pritamps (approved):
Approved!

PR #167, opened 2025-02-03
Updates sqs to hit backend

suryabulusu:
- Currently, SQS message is sent directly from the frontend
- Updating it to send SQS message from the backend. Frontend informs the backend to send an SQS message


src/services/API/sqs.js
```diff
@@ -1,32 +1,22 @@
-import { SQSClient, SendMessageCommand } from "@aws-sdk/client-sqs";
+import { client, dbClient } from "@/services/API/rootClient.js";
```
deepansh96: We should not really call the new backend as dbClient. Can you rename it to fastAPIClient? I know it doesn't change anything but just do a find and replace. Better for clarity

deepansh96: Also remove the aws sdk package from package.json as this is the only place we were using this

deepansh96: Remove the AWS_REGION, SQS_URL, ACCESS_KEY_ID and SECRET_KEY from the env.example and env.md. 
Also remove them from github workflow files

deepansh96: We should also remove sqs.js file as we wont need it anymore

suryabulusu: renamed to fastAPIClient. 

removed package

removed aws related env vars

retaining sqs js file -- coz thats where the connection to backend end point is made

PR #169, opened 2025-03-17
Adds AB Test v1 endpoints

suryabulusu:
Example URL: http://localhost:8080/?type=attendance&redirectTo=report&redirectId=FeedingIndiaStudents_67d53b5e0fcc46bf14d4e2b4&sub_type=reporting&group=FeedingIndiaStudents&auth_type=ID&abTestId=1

### DIFF
--- a/src/services/redirectToDestination.js
+++ b/src/services/redirectToDestination.js
@@ -63,10 +63,8 @@ export async function redirectToDestination(
       );
       let redirectURL = null;
       if (abTestResult.inTest) {
-        console.log(`User is in variant: ${abTestResult.variant}`);
         redirectURL = abTestResult.variantUrl;
       } else {
-        console.log(`User not in test: ${abTestResult.reason}`);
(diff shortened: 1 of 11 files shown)


src/services/redirectToDestination.js
```diff
@@ -54,7 +56,18 @@ export function redirectToDestination(
       break;
     }
     case "report": {
-      redirectURL = import.meta.env.VITE_APP_STUDENT_QUIZ_REPORT_BASE_URL;
+      const abTestResult = await abTestService.getABTestData(
```
pritamps, line 59: So we're assuming in frontend that AB test is for reporting?

Maybe it should be general and `report` should be part of AB test JSON?

pritamps (changes requested):
Just small change

src/services/redirectToDestination.js
suryabulusu: made an issue.. will do next pr 👍

